PR #39, opened 2020-02-03
Setting up an example for UITest with Espresso

tarekait1996:


app/src/main/res/layout/activity_main.xml
```diff
@@ -7,6 +7,7 @@
     tools:context=".MainActivity">
 
     <TextView
+        android:id="@+id/helloWorld"
```
Hambrsoom: why is it called helloWorld ??

PR #46, opened 2020-02-23
Created a bug report template

Hambrsoom:
I created a bug template that will be used whenever we open a new bug to stay consistent with our report structure


ferdousara-parvin:
Add another section that is optional to link the bug to a user story

Issue #21, opened 2020-01-14
Overlays for campus buildings


thomasgauvin:
As a user, I would like to be able to see which buildings are Concordia campus buildings, so that I can identify the buildings I have access to.

Acceptance criteria:
1. to see Concordia buildings over layered for distinguishing them from the other buildings.
2. the building's information once Clicking on an over layered building (Concordia Buildings)

Story tasks:
- [ ] Get all building overlay coordinates for Concordia campuses
- [ ] Draw overlays for each building
- [ ] Testing

PR #53, opened 2020-02-27
feature/ #21: overlay buildings

imabhijit:
This feature adds the overlay of polygons on the buildings that belong to Concordia University. The markers are clickable and clicking on it centers the building on the map, as well as return the building code that can be used for implementing new features that depend on building.


app/src/androidTest/java/com/example/concordia_campus_guide/LocationFragmentUITest.java
```diff
@@ -1,27 +1,31 @@
...
 
     @Test
-    public void dummyTest(){
-        onView(allOf(withId(R.id.MainActivityText), withText("Hello World!")));
+    public void polygonVisibleTest(){
+        onView(withId(R.id.mapView));
+        onView(withId(R.id.SGWBtn)).perform(click());
+        onView(withId(R.drawable.ev));
+        onView(withId(R.id.loyolaBtn)).perform(click());
+        onView(withId(R.drawable.ad));
```
tarekait1996: here, I do not think you are verifying much, you should do something like this: onView(withId(...)).check(matches(isDisplayed())). the matcher is important

app/src/main/java/com/example/concordia_campus_guide/BuildingCode.java
```diff
@@ -0,0 +1,68 @@
...
+    RR,
+    S,
+    SB,
+    T,
+    TD,
+    V,
+    VA,
+    X,
+    Z
+}
```
tarekait1996: https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3?gi=512c816544d0 please see this, enum is not very good as they consume a lot of resources, Android recommend using other types of structure such as TypeDef.

app/src/androidTest/java/com/example/concordia_campus_guide/LocationFragmentUITest.java
Hambrsoom: I fixed this and added a new UI test

app/src/main/java/com/example/concordia_campus_guide/BuildingCode.java
Hambrsoom: this modification would cost us a lot of time... I will keep it in mind and change it next sprint

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -1,31 +1,181 @@
...
+        return geoJsonPolygonStyle;
+    }
+
+
+    /**
+     * @param buildingCode it represents which building we will be covering
+     * @return Int of drawable resource's bitmap representation
+     */
+    public int getIcon(BuildingCode buildingCode){
+        switch (buildingCode){
```
thomasgauvin: Is there a way to programmatically get the resource, rather than having this huge switch statement? For example, I stored my images in assets, and I can access them this way: https://github.com/Concordia-Campus-Guide/Concordia-Campus-Guide/blob/5bfafbb19bdea5f2585bbc172a69bf097d812dc4/app/src/main/java/com/example/concordia_campus_guide/InfoCardFragment/InfoCardFragment.java#L70

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -130,30 +125,51 @@ public void onClick(View view) {
...
+    }
 
-        polygon.setStrokeWidth(ClassConstants.POLYGON_STROKE_WIDTH_PX);
-        polygon.setStrokeColor(ClassConstants.COLOR_BLACK_ARGB);
-        polygon.setFillColor(ClassConstants.COLOR_WHITE_ARGB);
+    public boolean setupMarkerClickListener(GoogleMap map) {
+        map.setOnMarkerClickListener(new GoogleMap.OnMarkerClickListener() {
+            @Override
+            public boolean onMarkerClick(Marker marker) {
+                //TODO: Make function that pops up the info card for the building (via the building-code)
```
thomasgauvin: We should list the issue # when we add a TODO comment

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
Hambrsoom: Why did you create an assets folder ? It is providing the same services as the res folder.

thomasgauvin: because res is not meant to be accessed programmatically, and I don't think it is good to create a very long switch statement to return the correct res

thomasgauvin: ```So why have an assets folder at all? If you want to compute the asset you want to use at run time, it's pretty easy. With resources, you would have to declare a list of all the resource IDs that might be used and compute an index into the the list. (This is kind of awkward and introduces opportunities for error if the set of resources changes in the development cycle.) (EDIT: you can retrieve a resource ID by name using getIdentifier, but this loses the benefits of compile-time checking.) Assets can also be organized into a folder hierarchy, which is not supported by resources. It's a different way of managing data. Although resources cover most of the cases, assets have their occasional use.```

https://stackoverflow.com/questions/5583608/difference-between-res-and-assets-directories

thomasgauvin: I also think it's useful to have subfolders, because too many files will easily get lost

thomasgauvin: coming back to my original question (we can continue the assets vs res discussion elsewhere), do you think it is good practice to create a very long switch in order to use the res?

Hambrsoom: Ok. that's a reasonable point. I am just curious. For your part, did you create a sub-folder in the assets ?

Hambrsoom: I agree with you. It is not. @tarekait1996  had a similar comment about that.

thomasgauvin: yes, for my part, I have a sub-folder in the assets `buildings_images`

imabhijit: Hi, the reason I had put the images in the res folder is that the map markers can only use bitmap images as icons, and that was the only option I could find to convert png images to bitmap easily. From what I looked up every res in the drawable folder will have a bitmap representation of it. Furthermore, that was also the way that was shown in the google map docs for adding custom icons.

PR #60, opened 2020-02-27
Added a template for possible improvements

Hambrsoom:
I created an improvement template that will be used whenever we open a new suggestion of improving the UI to stay consistent with our report structure


ferdousara-parvin:
Cool! Just change the name to "UI Improvement Template" so it is clear that it is a template for UI related things. Also why can't we create a template for improvements in general?

Hambrsoom:
It is possible to generalise it and make for any type of improvement such as performance improvement. Thus, in that case, I don't need to change the title though.

ferdousara-parvin:
Sounds good then just tweak the template so that it can be used for any kind of improvement.

PR #65, opened 2020-02-28
feature/#22: indoor details map

viveanban:
- The different floor maps are added as ground overlays of google maps.
- Native Indoor maps of google maps are disabled in order to use our custom level pickers.
- Custom Floor Picker is created as a grid view, with a custom adapter and listeners created to change the floor names for each building.


app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -112,8 +140,15 @@ public void onMapReady(GoogleMap googleMap) {
...
      */
+    private void initFloorPlans() {
+        hallGroundOverlay = mMap.addGroundOverlay(mViewModel.getHallBuildingOverlay());
+        mbGroundOverlay = mMap.addGroundOverlay(mViewModel.getMBBuildingOverlay());
+        vlGroundOverlay = mMap.addGroundOverlay(mViewModel.getVLBuildingOverlay());
+    }
+
     private void setMapStyle(GoogleMap googleMap) {
         try {
+            googleMap.setIndoorEnabled(false);
```
Hambrsoom: since we won't using the indoorMaps provided by google. Why don't we remove it from here ?

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -209,13 +262,26 @@ public boolean onMarkerClick(Marker marker) {
      * maps such as Current Location
      * @param mMap is the map used in the application
      */
+    public boolean setupClassMarkerClickListener(GoogleMap map) {
+        map.setOnMarkerClickListener(new GoogleMap.OnMarkerClickListener() {
+            @Override
+            public boolean onMarkerClick(Marker marker) {
+                System.out.println(marker.getTag());
```
Hambrsoom: Needs to change the usage of System.out.println to Log

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -245,6 +311,15 @@ private void getLocationPermission(){
         }
     }
 
+    private void classRoomCoordinateTool(GoogleMap map){
+        map.setOnMapClickListener(new GoogleMap.OnMapClickListener() {
+            @Override
+            public void onMapClick(LatLng latLng) {
+                System.out.println("\"coordinates\" : [" + latLng.longitude + ", " + latLng.latitude + "]");
```
Hambrsoom: change Systme.out.println to Log

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -254,13 +329,35 @@ private boolean requestPermission(){
...
+                mViewModel.setHallFloorplan(hallGroundOverlay, i==0?9:8);
+                break;
+            case MB:
+                mViewModel.setMBFloorplan(mbGroundOverlay, i==0?1:-1);
+                break;
+            case VL:
+                mViewModel.setVLFloorplan(vlGroundOverlay, i==0?2:1);
+                break;
+
+                default:
```
Hambrsoom: we can remove default from here since it is not doing anything no ?

app/src/main/AndroidManifest.xml
```diff
@@ -9,7 +9,7 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/AppTheme">
-        <activity android:name=".MainActivity">
+        <activity android:name=".Activities.MainActivity">
```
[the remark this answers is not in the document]
tarekait1996: it is a new folder to group all our activities together, I did the same for Adapters, Fragments and Interfaces

app/src/main/java/com/example/concordia_campus_guide/Adapters/FloorPickerAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+        return floorsAvailable.size();
+    }
+
+    @Override
+    public Object getItem(int i) {
+        return floorsAvailable.get(i);
+    }
+
+    @Override
+    public long getItemId(int i) {
```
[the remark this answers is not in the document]
tarekait1996: that's android auto-generated code but okay I guess I can change it

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
tarekait1996: from where? I am disabling it here...

Hambrsoom: In UiSetting() method in locationFragment

app/src/main/java/com/example/concordia_campus_guide/Adapters/FloorPickerAdapter.java
```diff
@@ -0,0 +1,77 @@
...
+        return i;
+    }
+
+    @Override
+    public View getView(final int i, final View view, ViewGroup viewGroup) {
+        Button button;
+        if (view == null) {
+            button = new Button(context);
+            button.setTextColor(Color.parseColor("#FFFFFF"));
```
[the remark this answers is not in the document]
thomasgauvin: *colors.xml right? just so we don't get confused

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -51,14 +54,126 @@ private GeoJsonLayer initLayer(GoogleMap map, Context applicationContext){
...
+    public void setMBFloorplan(GroundOverlay groundOverlay, int floorNb){
+        if(floorNb == -1)  setMBS2Floorplan(groundOverlay);
+        else if(floorNb == 1)  setMB1Floorplan(groundOverlay);
+    }
+
+    /**
+     * set ground overlay image for Mb S2 floor
+     * @param groundOverlay
+     */
+    private void setMBS2Floorplan(GroundOverlay groundOverlay){
```
thomasgauvin: I am concerned with this type of method. It seems every time we will want to add a floor, we will have to add a function in the code. This does not seem scalable as we add more floors. I think it would make sense to have a single parametrized method that sets a different image for the ground overlay based on the parameter. 

For example, for our info card, we added the building images using a single method, that took in a parameter of the building code and fetched the appropriate image. https://github.com/Concordia-Campus-Guide/Concordia-Campus-Guide/blob/6d2b4320216b184915428c51e7130f846ece6d7c/app/src/main/java/com/example/concordia_campus_guide/InfoCardFragment/InfoCardFragment.java#L91

In your case, you could have a setFloorplan(floorId) method that sets the floorplan according to a parameter passed to the function

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -254,13 +333,32 @@ private boolean requestPermission(){
         return (checkSelfPermission(getContext(), ClassConstants.FINE_LOCATION) == PackageManager.PERMISSION_GRANTED);
     }
 
+    @Override
+    public void onFloorPickerOnClick(int position, View view) {
+        if (selectedFloor != null) selectedFloor.setEnabled(true);
+        selectedFloor = (Button)view;
+        view.setEnabled(false);
+        switch(currentFloorPickerAdapter.getBuildingCode()){
```
thomasgauvin: I believe the same comment I made below applies here as well, why not have a single parametrized method that sets the floor plan according to the parameter passed? This will avoid us having to rewrite source code every time we want to add a floor/new building.

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -112,6 +140,12 @@ public void onMapReady(GoogleMap googleMap) {
      * The purpose of this method is to figure the style of the map to display
      * @param googleMap is the map that is used in the application
      */
+    private void initFloorPlans() {
```
thomasgauvin: Perhaps we could store these in a ground overlay array, so that our code can grow without needing to rewrite the code?

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -106,12 +239,32 @@ public GeoJsonPolygonStyle getPolygonStyle() {
         return geoJsonPolygonStyle;
     }
 
+    public ArrayList<String> getFloorsAvailable(BuildingCode buildingCode){
+        ArrayList<String> floorsAvailable = new ArrayList<>();
+        switch(buildingCode){
```
thomasgauvin: I believe the same comment I made previously applies here

tarekait1996:
Viveka and I refactored most of our previous code, please take a look now

app/src/test/java/com/example/concordia_campus_guide/LocationFragmentViewModelTest.java
```diff
@@ -44,4 +59,49 @@ public void getPolygonStyleTest(){
     public void getMapStyle(){
         assertEquals(viewModel.getMapStyle(), R.raw.mapstyle_retro);
     }
+
+//    @Test
```
thomasgauvin: Is there a reason why these tests are not deleted?

Issue #66, opened 2020-02-29
Unexpected buttons appearing when clicking on a building


tarekait1996:
**Describe the bug**
Two buttons show up on the screen when clicking on a building overlay. These redirect us to google map and should not be visible normally

**User Story Reference (Optional)**
#21 

**To Reproduce**
- Open the application
- Click on H building overlay
- on the bottom, you will see two small icons appearing

**Expected behavior**
The two buttons should not appear

**Screenshots**
![Screenshot_1582909177.png](https://images.zenhubusercontent.com/5e17476e0eda874c7740ff14/ecf470ac-fcf0-4d49-8113-7161d06e3077)

Hambrsoom:
Bug has been fixed :)

PR #68, opened 2020-02-29
#66/bug fixture Unexpected buttons appearing when clicking on a building

Hambrsoom:
I fixed the unexpected buttons appearing when clicking on a building by disabling the  map's tool bar setMapToolbarEnabled(false). i also fixed the issue of the improvement template which wasn't showing up in the list of templates.


.github/ISSUE_TEMPLATE/improvement-template-.md
```diff
@@ -1,12 +1,3 @@
----
-name: 'Improvement Template '
```
thomasgauvin: Why are we changing templates in this pull request? If they need to be changed, in the future, I think it would be better to keep our pull requests contain a single fix, so I would split this PR, 1 for the bug fix, one for the template change

Hambrsoom: I amnot changing the template. For some reasons, the template wasn't showing up in the template list because assign was empty. ok. that's a good point. will keep it in mind :)

PR #69, opened 2020-02-29
67: Removed the marker title

Hambrsoom:
I removed the title from the marker to not show it when someone clicks on the marker. This would fix the bug.


app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -88,7 +88,8 @@ private void addMarker(GoogleMap map, LatLng centerPos, BuildingCode buildingCod
                         .flat(true)
                         .anchor(0.5f,0.5f)
                         .alpha(0.90f)
-                        .title(buildingCode.toString())
+                        //This line should be included whenever we test the UI for the marker:
+                        //.title(buildingCode.toString())
```
tarekait1996: Is there really no other way to make the ui test without having to uncomment this line everytime ? If in the future we require to have similar behavior to test classes it might be a pain to uncomment all those future lines

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
```diff
@@ -256,7 +258,7 @@ private boolean requestPermission(){
 
     @Override
     public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
-
+        initMap();
```
viveanban: Why are we initializing the map at both line 261 and 66?

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
viveanban: I agree that this will be confusing in the future. Maybe mocking that data for the test would work?

Hambrsoom: It might be possible. However, at the moment, I didn't find another solution

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragment.java
Hambrsoom: I added it by mistake n forgot to remove it.

PR #71, opened 2020-02-29
feature/#21 : refactored the code for marker

Hambrsoom:
I deleted the usage of the enum class for building labels. Instead, I stored the pictures in assets and accessed them according to the file name. This will make it easier to add new buildings later.


app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -91,49 +94,69 @@ private void setPolygonStyle(GeoJsonLayer layer, GoogleMap map){
...
-                        .position(centerPos)
-                        .icon(BitmapDescriptorFactory.fromResource(R.drawable.h))
-                        .flat(true)
-                        .anchor(0.5f,0.5f)
-                        .alpha(0.0f)
-        );
-        marker.setTag(classNumber);
+    private BitmapDescriptor setupMarker(String buildingLabel, Context context) {
+        int height = 150;
+        int width = 150;
```
tarekait1996: would be good to have them in our constant class or somewhere else since they seem to be a constant/set attribute

app/src/test/java/com/example/concordia_campus_guide/LocationFragmentViewModelTest.java
```diff
@@ -32,20 +27,6 @@ public void init() {
         viewModel = new LocationFragmentViewModel();
     }
 
-    @Test
```
tarekait1996: are you going to introduce other tests in the future to test the new method ?

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -91,49 +94,69 @@ private void setPolygonStyle(GeoJsonLayer layer, GoogleMap map){
 
             String[] coordinates = feature.getProperty("center").split(", ");
             LatLng centerPos = new LatLng(parseDouble(coordinates[1]), parseDouble(coordinates[0]));
-            addBuildingMarker(map, centerPos, Enum.valueOf(BuildingCode.class, feature.getProperty("code")));
+            addBuildingMarker(map, centerPos, feature.getProperty("code"), context);
         }
```
tarekait1996: this is not specifically for you or your code but I noticed we used these methods a lot to get code properties such as code height, center, etc. maybe we should extract this in a method where we call it and get it directly, seems like duplicate codes. I thought we might want to have a building object created with all those info and then we just call the getters

PR #73, opened 2020-03-01
feature/#5: info card

ferdousara-parvin:
Info card is finally implemented! The only remaining part before closing the story is doing the appropriate test.


app/src/main/res/layout/info_card_fragment.xml
```diff
@@ -1,203 +1,186 @@
...
-                        android:layout_marginBottom="20dp"
-                        android:text="TextView" />
-                </LinearLayout>
-
-            </LinearLayout>
+                android:layout_height="wrap_content"
+                android:layout_marginTop="10dp"
+                android:gravity="center"
+                android:text="Services"
+                android:textColor="#932339"
```
tarekait1996: i think you should add these hex in a style inside color.xml since we might reuse them in the future

tarekait1996: you can even create styles with colors, text, and more all inside and just pass the style to a component, it helps with reusability

tarekait1996: this also applies to all the previous cases where you used a hex color directly

ferdousara-parvin: ok will do, thanks!

app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
```diff
@@ -0,0 +1,153 @@
+package com.example.concordia_campus_guide.Models;
+
+import android.text.TextUtils;
+
+import java.util.List;
+
+public class Building {
```
viveanban: Can all the info that will be displayed on the info card be wrapped as BuildingInfo? Just for the sake of making this more concise.

ferdousara-parvin: Thomas, Tarek and I decided on merging everything into one Building class actually.

PR #76, opened 2020-03-01
documentation/#64: Created Place, Building, Floor, Room, Direction, Route , etc

viveanban:
- Models are created

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
@@ -1,60 +1,48 @@
 package com.example.concordia_campus_guide.Models;
 
-
-import android.text.TextUtils;
-
 import android.text.TextUtils;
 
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
```diff
@@ -0,0 +1,86 @@
+package com.example.concordia_campus_guide.Models;
+
+import com.google.android.gms.maps.model.LatLng;
+
+import java.util.List;
+
+public class Building extends Place {
```
tarekait1996: just a heads up, this will need to be modified as Thomas and my changes added more functionality and attributes to this class

viveanban: I understand, I'll merge on top of the other changes before finalizing this PR.

tarekait1996:
please fix conflicts

PR #78, opened 2020-03-01
hotfix/#22/ fixed issue with floorpicker not appearing when clicking on marker

tarekait1996:
Fixed issue that we had when clicking on a marker and not seeing the floor picker. 
I had to switch some logic to use Buildings object instead of geoJSON for some method so that it could be used in other cases where the GeoJson is not available inside a method (basically storing the objects)


app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
```diff
@@ -1,27 +1,39 @@
...
-     * Constructor
-     */
-    public Building(){
-    }
+    private GroundOverlayOptions groundOverlayOption;
 
-    /**
-     * Constructor
-     */
+    public Building(LatLng centerCoordinates, String[] availableFloors, String buildingCode, float width, float height, float bearing) {
```
Hambrsoom: So, you and Thomas created the same class ?

app/src/main/java/com/example/concordia_campus_guide/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -64,23 +64,49 @@ private GeoJsonLayer initLayer(GoogleMap map, Context applicationContext){
...
+        float building_bearing = -1;
+        if (feature.getProperty("bearing") != null)
+            building_bearing = Float.parseFloat(feature.getProperty("bearing"));
+
+        String building_code = feature.getProperty("code");
+
+        return new Building(centerPos,floorsAvailable, building_code, building_width, building_height, building_bearing);
+    }
+
+    public LatLng getCenterPositionBuildingFromGeoJsonFeature(GeoJsonFeature feature){
```
viveanban: The mthd names could be shorter. We know from the paramerters that wtv info we're getting from is from a geoJsonFeature. I'd suggest a simple getBuildingCenter or computer parseBuildingCenter.

app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
tarekait1996: yes, I will fix conflicts

PR #85, opened 2020-03-02
#8/locate rooms 2

CarenRizk:
Added markers on each room for h_8, h_9, vl_1, vl_2, mb_1, mb_s2. The markers are invisible but are linked to an event listener, this way clicking on the text (classroom number) will be detected. The code might need some refactoring to add models instead of json files.


app/src/main/assets/buildingcoordinates.json
```diff
@@ -0,0 +1,1541 @@
+{
+  "type": "FeatureCollection",
```
thomasgauvin: do we need this file?

CarenRizk: It was used before for the polygons.. not sure if we still need it

Issue #86, opened 2020-03-04
Make the campuses buttons switch appear in the middle instead of left corner


tarekait1996:
As requested by the PO, the two buttons responsible for the switch between campuses should be displayed on the top-center of the screen instead of top-left corner.

PR #87, opened 2020-03-04
#86: fixed position of campus buttons

tarekait1996:
as required by PO, we centered the two campus switch buttons on screen.
Screenshot_1583296572

### DIFF
--- a/app/src/main/res/layout/location_fragment_fragment.xml
+++ b/app/src/main/res/layout/location_fragment_fragment.xml
@@ -10,32 +10,45 @@
         android:layout_height="match_parent"
         android:visibility="visible"
         tools:visibility="visible" />
+
     <LinearLayout
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:orientation="horizontal"
-        android:padding="10dp">
+        android:id="@+id/topMapContainer"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:gravity="center_horizontal"
+        android:orientation="vertical">
 
-        <Button
-            android:id="@+id/SGWBtn"
+        <LinearLayout
+            android:id="@+id/campusSwitchBtnContainer"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/location_fragment_fragment.xml
```diff
@@ -10,32 +10,45 @@
...
-            android:background="@color/colorAccent"
-            android:text="Loyola"
-            android:textColor="@color/textColor"></Button>
+            <Button
+                android:id="@+id/SGWBtn"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginRight="5dp"
+                android:layout_weight="1"
+                android:background="@color/colorAccent"
```
viveanban: Didn't the PO also request for the buttons to look less "disabled"? I'd suggest changing these button colours to @color/colorAppTheme.

thomasgauvin: i would do this in this pr if possible

thomasgauvin:
I would not do what @Hambrsoom  recommends because he did not ask that

app/src/main/res/layout/location_fragment_fragment.xml
tarekait1996: Screen Shot 2020-03-06 at 4 54 14 PM
I don't get it, how do they look disabled? what color should I put

tarekait1996: They are the same color in info card, so to stay consistent, I thought we keep them like this

tarekait1996: Screen Shot 2020-03-06 at 5 19 10 PM
is this good?

tarekait1996: @thomasgauvin @viveanban

Issue #88, opened 2020-03-04
Indicate default floor number on floor picker on first click


viveanban:
As requested by the PO, the floor picker should indicate the default floor when it first appears (top-most floor).

PR #89, opened 2020-03-04
#88: floor picker indicates top-most floor by default

viveanban:
As per the PO's request, the floor picker is selected at the top-most floor level by default.


app/src/main/java/com/example/concordia_campus_guide/Adapters/FloorPickerAdapter.java
```diff
@@ -51,13 +52,23 @@ public View getView(final int position, final View view, ViewGroup viewGroup) {
             button.setOnClickListener(new View.OnClickListener() {
                 @Override
                 public void onClick(View v) {
```
Hambrsoom: The only comment I have is the usage of "v" as a parameter name. It is preferred to use a clear parameter names such "view"

Issue #90, opened 2020-03-04
Zoom in On the User's current location 


Hambrsoom:
As requested by the PO, the map's view should display the user's current location instead of showing the SGW campus buildings on the map when opening the application for the first time.

PR #91, opened 2020-03-04
#90: Displaying the user location when lauching the application

Hambrsoom:
I modified the application to display the user's location when launching the application according to the PO's request. In case the user doesn't want to enable their current location, the map's view will be redirected towards the SGW campus. 

github

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="JDK" project-jdk-type="JavaSDK">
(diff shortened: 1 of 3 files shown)


.idea/misc.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="JDK" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
viveanban: Is this a needed change? or can it be ignored?

thomasgauvin:
Good to merge?

Issue #83, opened 2020-03-01
Research of Using Database in Android Applications


Hambrsoom:
As a developer, I would like to be able to research how to use SqlLite in Android Application, so that I can replace it instead of using JSON Files to store data.

Acceptance criteria:

Implementation plan.

Story tasks:

Research
Document
Knowledge Transfer (KT) with team

PR #96, opened 2020-03-08
#83: Added coordinates to Building class + db

thomasgauvin:
Added coordinates to Building class + db. You can now store coordinates into the building class in the db. I've also changed the json file for the new format


app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
```diff
@@ -39,10 +43,9 @@
     private float bearing;
 
     //Not sure what to do here yet
-    //
-    //
-    @Ignore
-    private List<List<List<Double>>> coordinates;
+    @Embedded
```
Hambrsoom: Did you try this? Because I read that u can't use @Embedded when you have the @Entity annotation on the class.

Hambrsoom: Unless I misunderstood something.

app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
```diff
@@ -72,7 +75,7 @@
     public Building(){}
     public Building(Double[] centerCoordinates, List<String> availableFloors, float width, float height, float bearing,
                     String campus, String buildingCode, String Building_Long_Name, String address,
-                    List<String> departments, List<String> services, List<List<List<Double>>> coordinates) {
+                    List<String> departments, List<String> services, ListOfCoordinates cornerCoordinates) {
         super(centerCoordinates);
```
Hambrsoom: I think we can start thinking about using the Builder design pattern over here maybe if it is applicable.

app/src/main/java/com/example/concordia_campus_guide/Models/ListOfCoordinates.java
```diff
@@ -0,0 +1,16 @@
+package com.example.concordia_campus_guide.Models;
+
+import java.util.List;
+
+public class ListOfCoordinates {
```
Hambrsoom: Did you create this class to be able to read from the geojson file for coordinates ?

app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
thomasgauvin: feel free to do so, maybe open a ticket for improvement, I don't think it should be done in this pr

app/src/main/java/com/example/concordia_campus_guide/Models/ListOfCoordinates.java
thomasgauvin: yes, the orm gave issues when we had List<Coordinates> in the Building class

app/src/main/java/com/example/concordia_campus_guide/Models/Building.java
thomasgauvin: it works when I run it (and check in debug), so unless you have a link saying its not good practice, I think it's good

Issue #98, opened 2020-03-12
Refactor Room Markers


CarenRizk:
Refactor #8 issue to use the database instead of the json files.

PR #111, opened 2020-03-14
#98/refactoring room markers

imabhijit:
Refactored the code for displaying the room markers to use the database to fetch rooms, and access them through the application state (cache).


imabhijit:
PR: https://github.com/Concordia-Campus-Guide/Concordia-Campus-Guide/pull/111

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -150,7 +149,7 @@ public void addBuildingMarker(GoogleMap map, LatLng centerPos, String buildingLa
                         .anchor(0.5f,0.5f)
                         .alpha(0.90f)
                         //This line should be included whenever we test the UI for the marker:
-                        //.title(buildingLabel)
+                        .title(buildingLabel)
```
Hambrsoom: Comment out this line here because otherwise a marker with the building label will show up everytime we click on the label

app/src/main/java/com/example/concordia_campus_guide/Models/Rooms.java
```diff
@@ -17,6 +21,47 @@ public Rooms(List<RoomModel> rooms) {
...
+    public List<RoomModel> getRoomsByFloor(String floor){
+        ArrayList<RoomModel> list = new ArrayList<>();
+        List<RoomModel> rooms = getRooms();
+        for(RoomModel room: rooms){
+            if(room.getFloorCode().equals(floor)){
+                list.add(room);
+            }
+        }
+        return list;
+    }
```
Hambrsoom: Wouldn't make more sense to have this method in the RoomModel class instead  ? because this class might be deleted later on in the future.

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
imabhijit: done

app/src/main/java/com/example/concordia_campus_guide/Models/Rooms.java
imabhijit: Hi, I put it in the Rooms class because when we initialize the app we import all the room data using a JSON and then store it in the rooms attribute of the Rooms object. Thus, this will allow us to bypass querying the database and simply loop through the list as needed.

Issue #95, opened 2020-03-08
Display Shuttle Routes


thomasgauvin:
As a user, I would like to be able to get the different routes taking the shuttle, so that I can easily know how to get to the other Concordia campus.

Acceptance criteria:
1. User can select shuttle as a transit option
2. User can select a time of departure, which is preset to the current time
3. User can see approximate time it takes for the shuttle
4. User can see a list of route options using the shuttle (say next 2 shuttles)

See pic below:
![image.png](https://images.zenhubusercontent.com/5e17455ee6c0ec60e23dc1a0/590abf81-0ab8-487a-9018-9fcba79b581f)

Story tasks:
1. Investigate Shuttle (API vs DB)
2. Get Routes from Shuttle API/DB
3. Communicate with GMap to have a common Route object
4. Display routes in list with time
5. Testing

PR #114, opened 2020-03-17
#95/shuttle data

CarenRizk:
Created the Shuttle model, JSON file for shuttle data, and UI to display the list of shuttles.


app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -53,6 +57,10 @@ protected void onCreate(Bundle savedInstanceState) {
         swipeableInfoCard = BottomSheetBehavior.from(infoCard);
 
         locationFragment = (LocationFragment) getSupportFragmentManager().findFragmentById(R.id.locationFragment);
+
+        AppDatabase appDB = AppDatabase.getInstance(this);
+        List<Shuttle> jas = appDB.shuttleDao().getScheduleByCampusAndDay("SGW", "Friday");
```
Hambrsoom: What is the purpose of this here ? Was it only for testing? if yes, can you delete it ? and if no, I don't think this a  good place for it. Also, we should use proper variable names.

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -53,6 +57,10 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        String gs = "Asda" ;
```
Hambrsoom: I don't think this line is doing anything here. right? If yes, can you remove it?

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -139,6 +147,12 @@ private void setUpDb(){
             Rooms rooms = ApplicationState.getInstance(this).getRooms();
             appDb.roomDao().insertAll(rooms.getRooms());
 
+            //Get shuttle schedule
+            Shuttles shuttles = ApplicationState.getInstance(this).getShuttles();
+            appDb.shuttleDao().insertAll(shuttles.getShuttles());
+
+            List<Shuttle> sh = appDb.shuttleDao().getScheduleByCampus("SGW");
```
Hambrsoom: Why are you retrieving the data for the shuttles in the SetUpDb() method? Is there a particular reason ? Because the  purpose of the setUpDb() method is just for inserting data inside the db.

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -39,4 +49,44 @@ public void setFrom(Place from) {
...
+            } else if (to.getClass() == Building.class) {
+                campusTo = ((Building) to).getCampus();
+            } else if (from.getClass() == Floor.class) {
+                campusFrom = ((Floor) from).getCampus();
+            } else if(to.getClass() == Floor.class) {
+                campusTo = ((Floor) to).getCampus();
+            }
+            if (campusFrom.compareTo(campusTo) == 0) {
+                return null;
+            }
```
Hambrsoom: Can you elaborate a bit more about this if else statements and what their purpose is ?

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -39,4 +49,44 @@ public void setFrom(Place from) {
...
+        }
+
+        Calendar cal = Calendar.getInstance();
```
Hambrsoom: Can you use calendar as variable name instead of cal? it would be clearer for the reader :)

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -39,4 +49,44 @@ public void setFrom(Place from) {
...
+
+    public String getShuttleDisplayText(List<Shuttle> shuttles) {
+        String content = "";
+        String campusTo = shuttles.get(0).getCampus().compareTo("SGW") == 0 ? "LOY" : "SGW";
+        if (shuttles == null || shuttles.size() == 0) {
+            return "No available routes using the shuttle service";
+        }
+        for (Shuttle shuttle : shuttles) {
+            content += shuttle.getCampus() + "  >   " + campusTo + ", leaves at: " + shuttle.getTime() + "\n";
+        }
```
Hambrsoom: I would put this a separate method and call it later in here. What do you think ? It would make it easier to test these two methods later.

app/src/main/java/com/example/concordia_campus_guide/Models/RoomModel.java
```diff
@@ -16,9 +16,7 @@
                         childColumns = {"floor_code"},
                         onDelete = CASCADE
                 ),
-                },
-        indices = {@Index(value = {"room_code"},
-        unique = true)})
+                })
```
Hambrsoom: Why did you remove the room_code as being unique ?? Were you having some issues ?

app/src/main/res/raw/buildingcoordinates.json
```diff
@@ -15,21 +15,24 @@
         "type": "Polygon",
         "coordinates": [
```
Hambrsoom: Buildingcoordinates is not used anymore. It will be deleted soon

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
CarenRizk: Yeah, it was only for testing. I'll remove it. Good catch!

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
CarenRizk: Because the from and to are places, so I can't get the campus since a place doesn't have campus. So I am checking if it is an instance of Floor or Building, then I am casting accordingly to be able to get the campus.

app/src/main/java/com/example/concordia_campus_guide/Models/RoomModel.java
CarenRizk: You were with me when we did this :P. Yes because "elevators", "womens" and "mens" are not unique.

app/src/test/java/com/example/concordia_campus_guide/LocationFragmentViewModelTest.java
```diff
@@ -4,6 +4,7 @@
 
 import com.example.concordia_campus_guide.Fragments.LocationFragment.LocationFragmentViewModel;
 import com.example.concordia_campus_guide.Models.Building;
+import com.example.concordia_campus_guide.Models.Shuttle;
```
tarekait1996: why did we add this import here?

app/src/main/res/layout/routes_activity.xml
```diff
@@ -70,15 +70,15 @@
...
                 android:layout_width="303dp"
                 android:layout_height="39dp"
                 android:layout_marginTop="8dp"
                 android:background="#29FFFFFF"
+                android:clickable="true"
                 android:ems="10"
                 android:inputType="textPersonName"
+                android:onClick="onClickFrom"
                 android:padding="8dp"
                 android:text="Name"
```
tarekait1996: Text should be written in strings.xml and referenced here, it will be easier that way when we support french text, we will just need to add the text in the french strings file with the same id.

app/src/main/java/com/example/concordia_campus_guide/Models/Floor.java
```diff
@@ -70,5 +73,9 @@ public void setAltitude(float altitude) {
     public String getDisplayName(){
         return floorCode;
     }
+
+    public String getCampus() { return campus; }
```
tarekait1996: This does not have to be in this PR specifically but please when updating/creating new files do not forget to update/create unit tests for the file. In this case, a floor unit test exists so maybe we should update it to test those methods?

app/src/main/java/com/example/concordia_campus_guide/Models/Campus.java
```diff
@@ -0,0 +1,5 @@
+package com.example.concordia_campus_guide.Models;
+
+public enum Campus {
```
tarekait1996: maybe we could use typedef, what do you think? by introducing more and more enums, it might slow our app. Look at this: https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3

app/src/main/java/com/example/concordia_campus_guide/Models/Floor.java
CarenRizk: Ok I updated the tests

app/src/main/java/com/example/concordia_campus_guide/Models/Campus.java
CarenRizk: Oh I'll delete this enum. I added it in the beginning and then decided to not use it. So now it's not being used anywhere

app/src/main/res/raw/buildingcoordinates.json
CarenRizk: ok..

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
CarenRizk: but the method is simple enough I don't think it needs to be more seperated

app/src/main/res/layout/routes_activity.xml
CarenRizk: image

CarenRizk: This is actually not part of what I did I'm not sure how the PR sees it as a new line.. But whoever worked on it is actually retrieving it by ID and replacing name so I think it should be fine

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
thomasgauvin: what about if the place is a room @CarenRizk ?

thomasgauvin: I see that floor and building both have a campus. we should also consider this for room. i think it makes sense to add campus to place

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -39,4 +49,44 @@ public void setFrom(Place from) {
...
+        shuttles = appDB.shuttleDao().getScheduleByCampusAndDayAndTime(campusFrom, day, time.format(calendar.getTime()));
+        return shuttles;
+    }
+
+    public String getShuttleDisplayText(List<Shuttle> shuttles) {
+        String content = "";
+        if (shuttles == null || shuttles.size() == 0) {
+            return "No available routes using the shuttle service";
+        }
+        String campusTo = shuttles.get(0).getCampus().compareTo("SGW") == 0 ? "LOY" : "SGW";
```
thomasgauvin: is this necessary? isn't getCampus() already returning "LOY" or "SGW"?

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -39,4 +49,44 @@ public void setFrom(Place from) {
...
+        String day = new SimpleDateFormat("EEEE", Locale.ENGLISH).format(calendar.getTime());
+        SimpleDateFormat time = new SimpleDateFormat("HH:mm");
+        shuttles = appDB.shuttleDao().getScheduleByCampusAndDayAndTime(campusFrom, day, time.format(calendar.getTime()));
+        return shuttles;
+    }
+
+    public String getShuttleDisplayText(List<Shuttle> shuttles) {
+        String content = "";
+        if (shuttles == null || shuttles.size() == 0) {
+            return "No available routes using the shuttle service";
```
thomasgauvin: might make sense to put this string in class constants

app/src/main/java/com/example/concordia_campus_guide/Database/Converters/StringLongConverter.java
```diff
@@ -0,0 +1,21 @@
+package com.example.concordia_campus_guide.Database.Converters;
+
+import androidx.room.TypeConverter;
+
+public class StringLongConverter {
```
thomasgauvin: is this converter for long types or time types? we should make it clear in the class name

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
CarenRizk: Yes but getCampus() is for the origin campus, this line is returning the destination campus. It's just for the UI

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
Hambrsoom: I agree with thomas of removing the campus attribute from all the subclasses and leave it only in the place class. However, we will need to modify the json file of the rooms accordingly to add the campus key and values in there for each room

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
thomasgauvin: agree with Caren here

app/src/main/java/com/example/concordia_campus_guide/Models/RoomModel.java
thomasgauvin: this will cause issues when we link walking points, we should discuss this

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
thomasgauvin: oh so it inverts it
I get it cool

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
Hambrsoom: You guys have a point. I am good with that too

app/src/main/java/com/example/concordia_campus_guide/Models/RoomModel.java
Hambrsoom: I agree with thomas cuz I noticed that in the json file we are storing elevators (as the roomCode). Which might be confusing when finding the paths using walkingpoints

Issue #115, opened 2020-03-18
create unit tests for model classes


tarekait1996:
Unit tests should be written to cover the logic inside our model classes. This include getter and setter as well as any other logic inside the code.

PR #116, opened 2020-03-18
#115: added unit tests and fixed jacoco and sonarqube issue

tarekait1996:
* Fixed SonarQube and Jacoco issue
* Excluded classes that should not be affecting the test code coverage
* Added 8 unit test classes for our models to reach 30% code coverage in the project
* updated some models to support testability


app/src/test/java/com/example/concordia_campus_guide/ModelTests/DirectionTest.java
```diff
@@ -0,0 +1,56 @@
+package com.example.concordia_campus_guide.ModelTests;
```
Hambrsoom: It looks good

Issue #17, opened 2020-01-13
Google Calendar Next Class Directions


thomasgauvin:
As a student, I would like to easily get directions to my next class as indicated in my Google Calendar, so that I can easily get to my next class.

Acceptance Criteria
Given that I gave access to my Google Calendar,
then, I should be able to:
- select the directions to my next class.
- get the locations and direction to my upcoming classes based on the next calendar event.

![image](https://user-images.githubusercontent.com/35609369/75820275-0d57dd00-5d6a-11ea-9871-36f4156a55bd.png)

narrapangan:
### Calendar Event Format 
This is the format that the user must follow when entering their classes on their calendar in order for the parsing to work correctly.

- Calendar Event Title
User must specify if it is a "Lecture: ", a "Tutorial: " or a "Lab: "
User must enter class name 
Eg. "Lecture: SOEN 357"

- Calendar Event Location
User must specify floor code and room code with a ", " in between
Eg. "H-9, 963"

PR #124, opened 2020-03-19
#17/google calendar next class directions

narrapangan:
Created the UI and added logic to get display next class and populate destination of the Route Activity with class location. Currently hidden behind the search results. To view and interact with the next class button, you will have to comment the search results out.


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Default" />
```
thomasgauvin: What is this class?

app/src/main/java/com/example/concordia_campus_guide/Activities/SearchActivity.java
```diff
@@ -66,6 +80,26 @@ public void onItemClick(AdapterView<?> parent, View view, int position, long id)
...
+        final CalendarEvent calendarEvent = calendarViewModel.getEvent(getApplicationContext(), this);
+
+        nextClassText.setText(calendarViewModel.getNextClassString(calendarEvent));
+
+        if(nextClassText.getText() != "No classes today"){
+            nextClassArrow.setVisibility(View.VISIBLE);
+            nextClassRow.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    openRoutesPage(new Place() {
```
thomasgauvin: This doesn't pass the actual place ie: with all attributes, maybe getting from the database, I think there should be a way to figure out how to pass the actual place as part of this story

app/src/main/java/com/example/concordia_campus_guide/Models/Helpers/CalendarViewModel.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final int PROJECTION_LOCATION_INDEX = 1;
+    private static final int PROJECTION_START_INDEX = 2;
+
+    public static final String[] INSTANCE_PROJECTION = new String[] {
+            Instances.TITLE,
+            Instances.EVENT_LOCATION,
+            Instances.DTSTART
+    };
+
+    public CalendarEvent getEvent(Context context, SearchActivity searchActivity) {
```
thomasgauvin: Since we are using context for many methods in this class, it might make sense to make a Context attribute in this class, which we instantiate in the constructor. This could be done by extending the AndroidViewModel class instead. Then we could rewrite all methods that require the Context as a parameter to access the context through the class attribute

app/src/main/res/layout/search_activity.xml
```diff
@@ -18,7 +18,7 @@
 
         <EditText
             android:id="@+id/searchText"
-            android:layout_width="357dp"
+            android:layout_width="0dp"
```
thomasgauvin: What is this change doing?

.idea/codeStyles/codeStyleConfig.xml
narrapangan: Whoops! I will delete this class

app/src/main/java/com/example/concordia_campus_guide/Models/Helpers/CalendarViewModel.java
narrapangan: That's a good point, I will do that.

app/src/main/res/layout/search_activity.xml
narrapangan: With a fixed width, the search bar text would be hidden by the back button. By using the constraints and keeping the width at 0dp, we ensure that this does not happen, no matter what phone will be displaying the search bar text.

app/src/main/java/com/example/concordia_campus_guide/Activities/SearchActivity.java
narrapangan: I will be passing a room (which is a Place), is that sufficient? Let me know!

thomasgauvin: yeah thats good, but i think it would be better to pass a classroom with the full information, rather than just the display name because we will need that

app/src/main/res/layout/search_activity.xml
thomasgauvin: nice!

Issue #38, opened 2020-02-03
Select indoor location


thomasgauvin:
As a user, I would like to select an indoor location so that I can get indoor directions to or from this location.

Acceptance criteria:
Given that I am using the indoor map of a building,then, I should be able to:
- select a room on the map as a location
- select a location as either an arrival location or a departure location for my routing needs
- select a departure location should give the user the option to get directions to or from this location

--

PR #126, opened 2020-03-19
#38 select location tests

thomasgauvin:
* Added unit tests for InfoCardViewModel, SearchActivityViewModel and RoutesActivityViewModel

* Added a way to decouple the ViewModels from the Context by using a ViewModelBuilder which builds ViewModels with the database, so that it is mockable

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/Helper/ViewModelFactory.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/Helper/ViewModelFactory.java
@@ -1,0 +1,36 @@
+package com.example.concordia_campus_guide.Helper;
+
+import android.app.Application;
+
+import androidx.lifecycle.ViewModel;
+import androidx.lifecycle.ViewModelProvider;
+
+import com.example.concordia_campus_guide.Activities.SearchActivityViewModel;
+import com.example.concordia_campus_guide.Database.AppDatabase;
+import com.example.concordia_campus_guide.Fragments.InfoCardFragment.InfoCardFragmentViewModel;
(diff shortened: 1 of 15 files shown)


app/src/test/java/com/example/concordia_campus_guide/ViewModel/InfoCardFragmentViewModelTest.java
```diff
@@ -0,0 +1,60 @@
+package com.example.concordia_campus_guide.ViewModel;
```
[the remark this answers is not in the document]
thomasgauvin: Yes, there is a way to mock the context of the application, but it seems complicated if we need the context within AppDatabase.getInstance(context)

thomasgauvin:
image

app/src/main/java/com/example/concordia_campus_guide/Helper/ViewModelFactory.java
```diff
@@ -0,0 +1,40 @@
...
+    private Application mApplication;
+    private AppDatabase appDb;
+
+
+    public ViewModelFactory(Application application) {
+        mApplication = application;
+        this.appDb = AppDatabase.getInstance(application.getApplicationContext());
+    }
+
+    //all view models need to be registered in the ViewModelFactory if they want to benefit from
```
ferdousara-parvin: put a javadoc instead?

thomasgauvin: sike

Issue #40, opened 2020-02-04
Show indoor directions from point X to Y between buildings


thomasgauvin:
As a user, I want to get indoor directions between two points that are located in separate buildings so that I can navigate through the campus.

Acceptance Criteria:
Given that I am using the indoor map of a building,
and then I choose my source point from one building,
and the I choose my destination point from another building,
and then I should be able to:
- get the routes from one building to another using the tunnels or other indoor passage.
- see the directions for the shortest path from point X (source) to point Y (destination) between 2 buildings.
-see the alternative paths and shortest paths are shown on the map.-
-save my preference of accessibility (wheelchair accessibility, no stairs, etc…).-
-see the pin of my current location and target destination across floors.-

viveanban:
Internal timeline: 6-13 March

PR #127, opened 2020-03-19
#40: Indoor directions through tunnels

viveanban:
- Added floors 00 and 1 to the Hall building
- Updated the walking point model to include a placeCode
- Changed the indoor search heuristic to consider different buildings
- Updates tests with new WalkingPoint ctor


app/src/main/java/com/example/concordia_campus_guide/Helper/IndoorPathHeuristic.java
```diff
@@ -12,23 +12,35 @@
 
     Context context;
 
-    public IndoorPathHeuristic(Context context){
+    public IndoorPathHeuristic(Context context) {
         this.context = context;
     }
```
Hambrsoom: @thomasgauvin  has created used factory design pattern to to create the view models that use the database. @thomasgauvin  do you think we can use it here too to avoid using context ?

app/src/main/java/com/example/concordia_campus_guide/Helper/IndoorPathHeuristic.java
```diff
@@ -12,23 +12,35 @@
...
 
     protected double computeHeuristic(final WalkingPoint currentCoordinate, final WalkingPoint destinationPoint) {
+
         if (!currentCoordinate.getFloorCode().equalsIgnoreCase(destinationPoint.getFloorCode())) {
-            final WalkingPoint accessPoint = getNearestAccessPointForFloor(currentCoordinate);
-            return getEuclideanDistance(currentCoordinate, accessPoint)
-                    + getEuclideanDistance(accessPoint, destinationPoint);
+            String currentBuildingCode = currentCoordinate.getFloorCode().split("-")[0];
+            String destinationBuildingCode = destinationPoint.getFloorCode().split("-")[0];
```
Hambrsoom: buildings don't have floor code. Thus, this might crash in that case. Did you test this case ?

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
```diff
@@ -57,21 +58,24 @@ public PathFinder(final Context context, final RoomModel source, final RoomModel
...
-        final WalkingPoint wantedPoint = new WalkingPoint(coordinates1, room.getFloorCode(), null, null);
+    protected WalkingPoint getWalkingPointCorrespondingToPlace(final Place place) {
+        String placeCode;
+        List<WalkingPoint> pointList = null;
+        if (place instanceof Building) {
+            placeCode = ((Building) place).getBuildingCode();
+            pointList = AppDatabase.getInstance(context).walkingPointDao().getAllWalkingPointsFromPlaceCode(placeCode);
+        } else if (place instanceof RoomModel) {
+            placeCode = ((RoomModel) place).getRoomCode();
+            pointList = AppDatabase.getInstance(context).walkingPointDao().getAllWalkingPointsFromPlaceCode(placeCode);
```
Hambrsoom: each of the extended classes (roomModel, Floor & building ) have different code attributes but they are all the same type and have the same purpose. Why don't we create a placeCode  in the place class  ? Then we won't need to have these if statements. What do you think ?Also, @thomasgauvin @tarekait1996  what do you guys think too?

app/src/main/java/com/example/concordia_campus_guide/Helper/IndoorPathHeuristic.java
thomasgauvin: Absolutely, this class can be rewritten using the viewmodelfactory because we only use the context to get the dao. So we could store the AppDatabase as an attribute to the IndoorPathHeuristic instead of the Context.

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
```diff
@@ -12,18 +15,16 @@
...
 
 /**
  * Usage Example:
- * Double[] src = {-73.57883681, 45.49731974};
- * Double[] trg = {-73.57873723, 45.49748425};
- * RoomModel room1 = new RoomModel(src,"H927" ,"H-9");
- * RoomModel room2 = new RoomModel(trg,"H842", "H-8");
- * PathFinder finder = new PathFinder(getContext(),  room1, room2);
- * List<WalkingPoint> solution = finder.getPathToDestination();
+ Double[] src = {-73.5791637, 45.49526596};
```
thomasgauvin: Are these still commented out?

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
thomasgauvin: Yes we should investigate this possibility

thomasgauvin: I will create an improvement task

thomasgauvin: https://github.com/Concordia-Campus-Guide/Concordia-Campus-Guide/issues/129

app/src/main/java/com/example/concordia_campus_guide/Models/WalkingPoint.java
```diff
@@ -48,11 +48,15 @@
...
-    public WalkingPoint(@NonNull Coordinates coordinate, @NonNull String floorCode, List<Integer> connectedPointsId, PointType pointType) {
+    @ColumnInfo(name = "place_code")
+    private String placeCode;
+
+    public WalkingPoint(@NonNull Coordinates coordinate, @NonNull String floorCode, List<Integer> connectedPointsId, PointType pointType, String placeCode) {
         this.coordinate = coordinate;
         this.floorCode = floorCode;
         this.connectedPointsId = connectedPointsId;
         this.pointType = pointType;
+        this.placeCode = placeCode;
```
thomasgauvin: Why do we need both placeCode and floorCode? We should be able to access the floor from the place no?

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
thomasgauvin: I see they also introduced a place code attribute, which I don't exactly understand how its being used. If it's being used the way you speak of above @Hambrsoom , we should do this as part of this story

Hambrsoom: As you mentioned before, this is not something related to this user story specifically. placeCode is like an id for the each type of place. That's why i thought we could have put it in the father class instead of having it in each sub class

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
viveanban: Yes, they are just an example usage for people to get an example fast fast.

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
viveanban: yes the placCode maps to either the building code, or the roomCodes of the places in order to get the place corresponding to the walkingpoint.

app/src/main/java/com/example/concordia_campus_guide/Models/WalkingPoint.java
viveanban: no, the place code is the id that maps the walkingpoint to a place. for rooms, it would return the roomCode (e.g 937) and for buildings it would return the buildingCode (e.g H-9). They are not combined as the name would suggest. I understand that this might be confusing, so do you have any other suggestions for the field's name?

app/src/main/java/com/example/concordia_campus_guide/Helper/IndoorPathHeuristic.java
viveanban: Good catch! I will look into it.

viveanban: After looking into it, you see that the currentCoordinate and desitinationPoint are actually WalkingPoints. Every walkingPoint has a floorcode so this is not an issue :)

app/src/main/java/com/example/concordia_campus_guide/Helper/IndoorPathHeuristic.java
viveanban: Instead of using a factory to do so, I think passing the database directly as a parameter is a better solution for this case. I will update this PR with my changes for that!

app/src/main/java/com/example/concordia_campus_guide/Helper/IndoorPathHeuristic.java
```diff
@@ -10,36 +10,49 @@
...
+    public IndoorPathHeuristic(final AppDatabase appDatabase) {
+        this.appDatabase = appDatabase;
     }
 
     /**
      * The heuristic is based on the distance between the goal and destination coordinates, the further a point from a destination, the bigger the heuristic it will have.
-     * @param currentCoordinate
+     * @param currentPoint
      * @param destinationPoint
      * @return
```
Hambrsoom: Can you add the description to the parameters as well please ?

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
```diff
@@ -34,17 +35,19 @@
...
 
-        final List<WalkingPoint> walkingPoints = AppDatabase.getInstance(context).walkingPointDao().getAll();
+        final List<WalkingPoint> walkingPoints = appDatabase.walkingPointDao().getAll();
         walkingPointNodesMap = populateWalkingPointMap(walkingPoints);
 
-        this.initialPoint = getWalkingPointCorrespondingToRoom(source, walkingPoints);
-        this.destinationPoint = getWalkingPointCorrespondingToRoom(destination, walkingPoints);
+        this.initialPoint = getWalkingPointCorrespondingToPlace(source);
+        this.destinationPoint = getWalkingPointCorrespondingToPlace(destination);
     }
```
Hambrsoom: This will make writing the unit tests way easier now 👍

app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
```diff
@@ -60,26 +63,24 @@ public PathFinder(final Context context, final RoomModel source, final RoomModel
         return walkingPointNodesMap;
     }
 
-    /**
-     * This method's purpose is to get a corresponding walking point given a room object.
-     * @param room
-     * @param walkingPointList
-     * @return
-     */
```
Hambrsoom: Why did you take of the javac doc ?

Issue #94, opened 2020-03-08
Display Routes of Google Maps + UI


thomasgauvin:
As a user, I would like to see my different route options for outdoor directions according to my transit preferences so that I can easily navigate from one building to another.

More details: This step includes calling the GMap API and getting the different routes available. This story also includes displaying the list of the different routes, with a time that it takes to complete the route.

See pic below:
![image.png](https://images.zenhubusercontent.com/5e17455ee6c0ec60e23dc1a0/a45d4913-51cd-4b80-b5b9-92976f3d62b2)

Acceptance criteria:
Given that I am a user, 
then, i open the application,
then, i search for the direction between two places,
then i should be able to:
- select his transit preferences
- get a list of possible routes
- see the time that each route takes

Steps:
1- Research GMap API
2- Implement transit buttons
3- Get routes from GMap
4- Display routes on Routes page
5- Testing

imabhijit:
Hi, it's not working for me when I try to select the location, I have PMed you the error code please take a look at it. 
Steps to reproduce:
1) I clicked on the search button
2) Select Hall F. Building from the dropdown
Then it crashes

imabhijit:
I also can't see the UI for selecting different transit types, is that expected behaviour?

imabhijit:
I also cannot see any routes yet, but I am guessing that it is currently the expected behavior?

PR #128, opened 2020-03-19
#94/display gmaps routes 3

ferdousara-parvin:
What I was able to achieve for this PR: Create a DirectionsResult object in the RoutesActivityViewModel which contains all the necessary information related to the possible routes from an origin to a destination with a provided transport type.

Even though user story #94 is not done (UI part left), a good chunk is done so I thought I would merge it as some of you guys will need the information retrieved from the Google Maps Directions API soon. 

Important notes:

- Don't be scared by the number of files changed, most of them are classes from Google Maps Services library so you can disregard them.

- The important files you should review are: DirectionsApiDataParser, DirectionsApiDataRetrieval, RoutesActivity, ClassConstants and RoutesActivityViewModel files.

- There's a folder called GoogleMapsServicesTools. DO NOT change anything in these files. The two folders inside are essentially a duplicate of "model" and "internal" folder in the Google Maps Services library. I needed to create this duplication because or else the project wasn't building properly.

- Instead of using the Google Maps services library, I had to import it into our own code to be able to modify some of the models that I needed from the library because the mapping from the JSON string returned by the API and the models provided by Google maps were not identical so Gson was not working properly.


thomasgauvin:
@ferdousara-parvin could you add a comment indicating which files were mostly autogenerated to save time for the pr reviewers?

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
```diff
@@ -76,6 +94,20 @@ public boolean onKeyDown(int keyCode, KeyEvent event)  {
         return super.onKeyDown(keyCode, event);
     }
 
+    public void directionsApiCallBack(DirectionsResult result)
+    {
+        mViewModel.setDirectionsResult(result);
+        System.out.println("lol");
```
thomasgauvin: @ferdousara-parvin

thomasgauvin: :')

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -35,4 +40,39 @@ public void setFrom(Place from) {
...
+     * Build the URL that will be used to call the Google Maps Directions API by passing the necessary parameters
+     *
+     * @param from: latitude and longitude of the origin
+     * @param to: latitude and longitude of the destination
+     * @param transportType: main transport type to get from origin to destination
+     *
+     */
+    public String buildUrl(LatLng from, LatLng to, @ClassConstants.TransportType String transportType) {
+        // Origin of route
+        String str_origin = "origin=" + from.latitude + "," + from.longitude;
```
thomasgauvin: I think all these comments are useless, the naming of your variables is good enough

thomasgauvin:
Hey Ferdou,

Since this PR is quite big, could you actually indicate which files the reviewers should look at? It is difficult to browse the files and know which ones need reviewing

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
ferdousara-parvin: smh thanks!

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
ferdousara-parvin: you're right!

ferdousara-parvin:
Also, I am aware that my commit messages don't follow the format we agreed on (having the issue number at the beginning of the commit message). I'll be more mindful of that in my next commits! Sorry!

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
```diff
@@ -96,6 +126,17 @@ private void openSearchPage(){
...
+        Coordinates toCoordinates = (mViewModel.getTo().getCenterCoordinates());
+        to = new MarkerOptions().position(new LatLng(toCoordinates.getLatitude(), toCoordinates.getLongitude()));
+
+    }
+
+    private void setTo() {
+        Coordinates fromCoordinates = (mViewModel.getFrom().getCenterCoordinates());
+        from = new MarkerOptions().position(new LatLng(fromCoordinates.getLatitude(), fromCoordinates.getLongitude()));
+    }
+
```
CarenRizk: Maybe I don't understand this right.. but why in the setFrom() you are setting the to, and vice versa for the setTo() ?

ferdousara-parvin: oh oups I made a mistake here, thanks for catching it :)

Issue #130, opened 2020-03-19
Refactor room numbers to be unique


CarenRizk:
Change the information in the json file for room numbers to make them unique. This way we can link a room to a walking point.

PR #133, opened 2020-03-19
#130: Unique room-codes

CarenRizk:
added click listener to room markers, updated roomCodes to be unique in room json


app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
```diff
@@ -310,7 +311,7 @@ public boolean setupClassMarkerClickListener(GoogleMap map) {
         map.setOnMarkerClickListener(new GoogleMap.OnMarkerClickListener() {
             @Override
             public boolean onMarkerClick(Marker marker) {
-                Log.i(TAG,marker.getTag().toString());
+                System.out.println(marker.getPosition());
```
thomasgauvin: Do we need to print this? Can we just remove this line?

Hambrsoom: We are never supposed to be using system.out.println in the application

PR #140, opened 2020-03-22
#130: unique room numbers

CarenRizk:
fixed some nonunique room codes and added a unique identifier to the room's model

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
@@ -311,7 +311,8 @@ public class LocationFragment extends Fragment implements OnFloorPickerOnClickLi
         map.setOnMarkerClickListener(new GoogleMap.OnMarkerClickListener() {
             @Override
             public boolean onMarkerClick(Marker marker) {
-                Log.i(TAG,marker.getTag().toString());
+                System.out.println(marker.getPosition());
+                // Log.i(TAG,marker.getTag().toString());
                 return false;
             }
         });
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
```diff
@@ -311,7 +311,8 @@ public boolean setupClassMarkerClickListener(GoogleMap map) {
         map.setOnMarkerClickListener(new GoogleMap.OnMarkerClickListener() {
             @Override
             public boolean onMarkerClick(Marker marker) {
-                Log.i(TAG,marker.getTag().toString());
+                System.out.println(marker.getPosition());
```
viveanban: Don't we prefer logs over system out prints?

app/src/main/res/raw/rooms_info.json
```diff
@@ -297,111 +297,111 @@
...
+      "roomCode":"842",
       "floorCode":"H-8",
       "centerCoordinates":{
         "longitude":-73.5789343714714,
         "latitude":45.49737026836251
       }
     },
     {
-      "roomCode":"elevator",
+      "roomCode":"elevators-2",
```
viveanban: Are we not distinguishing staff elevators from elevators accessible to everyone?

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
CarenRizk: yeah I forgot to remove it :s, I'll change it now

app/src/main/res/raw/rooms_info.json
CarenRizk: Fixed to have two different types, "elevators-#" and "staff elevators-#"

PR #142, opened 2020-03-22
#40/walking points

viveanban:
Updated the walking points to include more options.


narrapangan:
wow! more walking points! thanks for adding these!

ferdousara-parvin:
That's crazy!

Issue #11, opened 2020-01-12
Show me nearest points of interest of a type


tarekait1996:
As a user, I would like to see nearby points of interest according to the type I specified so that I can find the type of location that I am looking for, without needing prior knowledge of the existence of the place.

Acceptance Criteria:
1. The user should be able to choose a type of points of interest from a list of options (washroom, coffee shop, elevator, etc.)
2. The user should see a list of points according to his chosen type on the map
3. The user should be able to select a point and obtain directions
4. The nearest point can be an indoor or outdoor location.
5. The nearest point is highlighted somehow (numbering, centering of screen, etc.)

tarekait1996:
![Screen Shot 2020-03-23 at 7 04 04 PM](https://user-images.githubusercontent.com/34817695/77371192-22a4a380-6d39-11ea-95ee-a7737f136220.png)

PR #145, opened 2020-03-25
#11/ Point of interest UI

tarekait1996:
*NOT FINAL*
To keep my PRs small, I will be making one regularly (every day).
This one contains the following:
* A new fragment that will hold the point of interests logic
* A viewPagerAdapter as well as a GridView adapter to implement the UI of point of interests
* New icons for the point of interest. So far we have one for washrooms, escalators, elevators, lounges and water fountains 
Screenshot_1585097412


app/src/main/java/com/example/concordia_campus_guide/Adapters/PointOfInterestGvAdapter.java
```diff
@@ -0,0 +1,68 @@
...
+import android.widget.TextView;
+
+import com.example.concordia_campus_guide.R;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+class PointOfInterestGvAdapter extends BaseAdapter {
+    private Context context;
```
Hambrsoom: Is there a way to avoid using the context ?

app/src/main/java/com/example/concordia_campus_guide/Adapters/PointOfInterestGvAdapter.java
```diff
@@ -0,0 +1,68 @@
...
+        View view = LayoutInflater.from(context).inflate(R.layout.point_of_interest_item, parent, false);
+
+        TextView serviceTxt = view.findViewById(R.id.serviceTv);
+        ImageView serviceIv = view.findViewById(R.id.serviceIv);
+
+        serviceIv.setImageBitmap(getBitmapFromAssets("point_of_interest_icons/poi_"+services.get(position)+".png"));
+        serviceTxt.setText(services.get(position).replace("_", " "));
+
+        return view;
+    }
```
Hambrsoom: This class doesn't seem to be a ui class. why do you have this method here ?? can we place it somewhere else to avoid using the context attribute ?

app/src/main/java/com/example/concordia_campus_guide/Adapters/PointOfInterestGvAdapter.java
```diff
@@ -0,0 +1,68 @@
...
+    private Bitmap getBitmapFromAssets(String fileName){
+        AssetManager assetManager = context.getAssets();
+        InputStream inputStream = null;
+        try{
+            inputStream = assetManager.open(fileName);
```
Hambrsoom: u need to close() inputStream

app/src/main/java/com/example/concordia_campus_guide/Adapters/PointOfInterestVPAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.List;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+
+public class PointOfInterestVPAdapter extends RecyclerView.Adapter<PointOfInterestVPAdapter.MyViewHolder> {
+
+    private List<String> services;
+    private Context context;
```
Hambrsoom: is there a possibility to avoid using the context again here ?

app/src/main/java/com/example/concordia_campus_guide/Fragments/POIFragment/POIFragment.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private void initComponents(View rootView) {
+        poiVP = rootView.findViewById(R.id.poiVp);
+        setupViewPager();
+    }
+
+    private void setupViewPager() {
+        List<String> services = new ArrayList<>();
+        //TESTING PURPOSE, WILL BE REMOVED IN NEXT PR
```
Hambrsoom: Make sure you don't leave this comment

app/src/main/java/com/example/concordia_campus_guide/Fragments/POIFragment/POIFragmentViewModel.java
```diff
@@ -0,0 +1,6 @@
+package com.example.concordia_campus_guide.Fragments.POIFragment;
+
+import androidx.lifecycle.ViewModel;
+
+public class POIFragmentViewModel extends ViewModel {
```
Hambrsoom: when you start implementing this class use the factory method to create an object of this class. to avoid using context. check up the info card fragment view model as a reference.

viveanban: Yes it will be used in the next PR for this task.

app/src/main/java/com/example/concordia_campus_guide/Adapters/PointOfInterestGvAdapter.java
tarekait1996: I use it to link the layout the adapter will use, I do not see another way...

app/src/main/java/com/example/concordia_campus_guide/Adapters/PointOfInterestGvAdapter.java
tarekait1996: this needs to be here it is a required method for an adapter. How would the adapter know what view to return otherwise?

app/src/main/java/com/example/concordia_campus_guide/Fragments/POIFragment/POIFragment.java
tarekait1996: yes, I put it here so you guys understand this is not final

app/src/main/java/com/example/concordia_campus_guide/Fragments/POIFragment/POIFragmentViewModel.java
tarekait1996: ok I ll keep it in mind

app/src/main/java/com/example/concordia_campus_guide/Adapters/PointOfInterestGvAdapter.java
Hambrsoom: ooo i see... that makes sense .

app/src/main/java/com/example/concordia_campus_guide/Fragments/POIFragment/POIFragment.java
Hambrsoom: No problem with that 👍

PR #146, opened 2020-03-25
#11: Implementing what happens when clicked on poi

viveanban:
*WIP*
- Created a onClick listener to handle what happens when the POI is clicked
- Created an annotation (TypeDef) to define the POI types.
- Renamed the png files to reflect the POI types

_If you want to test this, you can replace the LocationFragment in the main activity with the POIFragment_


app/src/main/java/com/example/concordia_campus_guide/Models/PoiType.java
```diff
@@ -0,0 +1,10 @@
+package com.example.concordia_campus_guide.Models;
+
+//TODO: Make sure these strings correspond exactly to the PointTypes
```
viveanban: Before completing this, we will need to change the PointType into a typeDef ([for better performance](https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3)). The todo is created as a reminder (an improvement issue will probably be created too).

ferdousara-parvin: Yeah create an improvement issue asap in order to not forget

ferdousara-parvin:
Tested it, works great! :)

app/src/main/java/com/example/concordia_campus_guide/Models/PoiType.java
viveanban: Done! [#147]

PR #148, opened 2020-03-25
#11/poi implementation

tarekait1996:
This PR serves to finalize the UI part of us # 11.

WHAT WAS DONE: 

- A new Gridview called POIGridView, this one is based on a stack overflow suggestion due to the basic grid view not adapting dynamically to the right height. Therefore, if we kept the basic grid view we would have issues with the UI not displaying as it should
- I modified the logic for the swipeable card to reuse what @ferdousara-parvin previously implemented with the info card fragment card.
- I added dots indicator for the view pager to indicate the different pages we have in POI

WHAT IS LEFT:

- logic code to display the point of interests on the map after a click
- Allowing user to see a route from two points (including a point of interest)

VISUAL: 

ezgif com-optimize


app/src/main/java/com/example/concordia_campus_guide/Helper/POIGridView.java
```diff
@@ -0,0 +1,39 @@
...
+    public POIGridView(Context context, AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    public POIGridView(Context context, AttributeSet attrs, int defStyle) {
+        super(context, attrs, defStyle);
+    }
+
+    @Override
+    protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
```
thomasgauvin: Can you add a comment for this function? I don't understand what it's needed for

thomasgauvin:
I think your screenshot/screen recording is not working?

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -92,35 +88,51 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        fragmentTransaction.add(R.id.bottom_card_frame, poiFragment);
         fragmentTransaction.commit();
+
+        swipeableInfoCard.setState(BottomSheetBehavior.STATE_COLLAPSED);
+        swipeableInfoCard.setPeekHeight(180);
+        swipeableInfoCard.setHideable(false);
+
+    }
+
+    public void resetBottomCard(){
```
viveanban: I'd suggest changing the content of this method to only remove fragments in the bottom card (as per our discussion) instead of removing all fragments that are not LocationFragment. This can be done by creating a list that can contain all the fragments that will be put in the bottom card or a list that contains the frangment classes that are added to bottom cards.

viveanban:
I think we should either create an improvement task to move the zoom/unzoom buttons because right now they get hidden behind the cards. Thoughts?
image

app/src/main/java/com/example/concordia_campus_guide/Helper/POIGridView.java
tarekait1996: done

ferdousara-parvin:
Looks awesome!

Issue #49, opened 2020-02-26
Notification


thomasgauvin:
As a student, I would like to be notified of my upcoming class an hour in advance, so that I can get to class on time.

Acceptance criteria:
1. Pop-up shows up in the application 1h before class.
2. Pop-up has information about class

Story tasks: 
1. Create UI
2. Synchronize with calendar
3. Test

PR #165, opened 2020-03-26
#49/notification

Hambrsoom:
We have implemented the Notification user story where the user would get a pop when using the application which would tell them that they have a class soon as shown in the below screen shot.

image

image

image

Furthermore, this pop up will starting showing up every 15 minutes if the user is using the application. The pop up is working currently on the MainActivity at the moment. 

Finally, we also included some unit tests covering most of what we implemented. We couldn't write the unit tests for some of the methods due to the usage of system current time.


app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -53,11 +68,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
-
         locationFragment = (LocationFragment) getSupportFragmentManager().findFragmentById(R.id.locationFragment);
     }
 
+    public void popUp(CalendarEvent calendarEvent){
+        AlertDialog.Builder builder = new AlertDialog.Builder(MainActivity.this);
+        builder.setCancelable(true);
+        builder.setTitle("Heads up... Your next class is in " +  mViewModel.displayTimeToNextClass(calendarEvent.getStartTime()));
+        builder.setMessage("You have " + calendarEvent.getTitle() + " in " + mViewModel.displayTimeToNextClass(calendarEvent.getStartTime())  + " at " +
+                calendarEvent.getLocation() +" ! Please choose to either get directions to your class or to disregard this message");
```
thomasgauvin: Should rename disregard -> ignore
```suggestion
                calendarEvent.getLocation() +" ! Please choose to either get directions to your class or to ignore this message");
```

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -53,11 +68,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+        setupCancelBtn(builder);
```
thomasgauvin: Rename Cancel to Ignore

thomasgauvin:
You mention this popup will pop up every 15 minutes. Is this when the user chooses to ignore the pop up?

app/src/test/java/com/example/concordia_campus_guide/ViewModel/MainActivityViewModelTest.java
```diff
@@ -0,0 +1,44 @@
...
+import static junit.framework.TestCase.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class MainActivityViewModelTest {
+    private MainActivityViewModel mainActivityViewModel;
+
+    @Before
+    public void init(){
+        MockitoAnnotations.initMocks(this);
```
thomasgauvin: Why do you need this?

app/src/test/java/com/example/concordia_campus_guide/ViewModel/MainActivityViewModelTest.java
```diff
@@ -0,0 +1,44 @@
...
+        mainActivityViewModel = new MainActivityViewModel();
+    }
+
+    @Test
+    public void cleanTimeDisplayLongerThanOneHoursTest(){
+        long date = 15852069;
+        String dateString = mainActivityViewModel.timeUntilMethod(date);
+        String expectedValue ="04 hours and 24 minutes";
+        assertEquals("cleanTimeDisplayLongerThanOneHoursTest(): ", expectedValue, mainActivityViewModel.cleanTimeDisplay(dateString));
```
thomasgauvin: Can you make the message more descriptive? It doesn't have to match the method signature

app/src/test/java/com/example/concordia_campus_guide/HelpersTest/NotificationTest.java
```diff
@@ -0,0 +1,60 @@
...
+        );
+    }
+    @Test
+    public void getRoomTest(){
+        RoomModel expectedRoom = new RoomModel(new Coordinates(0, 0), "111", "H-9");
+        assertEquals("getRoomTest(): ", expectedRoom.getFloorCode(), notification.getRoom("H-927").getFloorCode());
+    }
+
+    @Test
+    public void getRoom10thFloorTest(){
```
thomasgauvin: Is this test necessary? isn't it the same as the test above?

app/src/main/java/com/example/concordia_campus_guide/Helper/Notification.java
```diff
@@ -0,0 +1,62 @@
...
+public class Notification {
+    private MainActivity mainActivity;
+    private AppDatabase appDatabase;
+
+    public Notification(MainActivity mainActivity, AppDatabase appDatabase){
+        this.mainActivity =  mainActivity;
+        this.appDatabase  = appDatabase;
+    }
+
+    public void checkUpCalendarEvery5Minutes() {
```
thomasgauvin: Can you reduce this time to check calendar? Make it like every 10 seconds

thomasgauvin: This should be in a global state which we refresh every 10 seconds because both the notification and the search uses it

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -53,11 +68,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        });
+    }
+
+    private void setupShowMeDirectionsBtn(AlertDialog.Builder builder, String location){
+        final String locationTemp = location;
+        final Activity mainActivity = this;
+        builder.setPositiveButton("Show me Directions", new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                RoomModel room = notification.getRoom(locationTemp);
```
thomasgauvin: Can you add a check to make sure that the event is properly stored in the Calendar? When the location is not properly stored, it causes the application to crash

app/src/test/java/com/example/concordia_campus_guide/ViewModel/MainActivityViewModelTest.java
imabhijit: that's true it can be removed if you aren't using any mocks.

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivityViewModel.java
```diff
@@ -2,13 +2,35 @@
...
+    public String displayTimeToNextClass(String date){
+        Date eventDate = new Date((Long.parseLong(date)));
+        long differenceInMillis = eventDate.getTime() - System.currentTimeMillis();
+        String timeUntil = timeUntilMethod(differenceInMillis);
+        return cleanTimeDisplay(timeUntil);
+    }
 
-    public MainActivityViewModel(@NonNull Application application) {
-        super(application);
+    public String cleanTimeDisplay(String timeUntil){
```
imabhijit: Maybe call this cleanTimeFormatString?

app/src/main/java/com/example/concordia_campus_guide/Helper/Notification.java
imabhijit: hmm yeah I agree but keep the pop up to 5mins I think it would be annoying if the popup came up every 10 seconds

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
Hambrsoom: done!

app/src/test/java/com/example/concordia_campus_guide/ViewModel/MainActivityViewModelTest.java
Hambrsoom: True, I had there at the beginning thinking i will need to mock something. Good catch guys

app/src/test/java/com/example/concordia_campus_guide/HelpersTest/NotificationTest.java
Hambrsoom: it is very similar to the other test. However, it covers different condition in the method. Thus, i created a separate method for each condition

app/src/main/java/com/example/concordia_campus_guide/Helper/Notification.java
Hambrsoom: I feel like 10 seconds is too short. Maybe we can reduce it to 5 minutes the check up if guys want. I will do this in a different pr though.

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
Hambrsoom: Should avoid displaying the event if the location is not properly stored or how should i fix this ?
Any suggestion?

thomasgauvin: Yes, just have a message that says that the location format is wrong in the notification

app/src/main/java/com/example/concordia_campus_guide/Helper/Notification.java
thomasgauvin: We should do this in this PR because this feature is buggy and I don't think we should merge buggy features to master

thomasgauvin: Ok I think it works

app/src/main/java/com/example/concordia_campus_guide/Database/Daos/RoomDao.java
```diff
@@ -28,6 +28,10 @@
     @Query("SELECT * FROM rooms WHERE id=:roomId")
     public RoomModel getRoomById(long roomId);
 
+    @Query("SELECT * FROM rooms WHERE room_code=:roomCode AND floor_code=:floorCode" )
+    public RoomModel getRoomByRoomCodeAndFloorCode(String roomCode, String floorCode);
+
+
```
viveanban: You created a method that already exists. getRoomByIdAndFloorCode takes the same parameters and fetches the same data.

Hambrsoom: oo I didn't noticed that. thank you I will check it out

Issue #154, opened 2020-03-26
Re-position/Remove zoom buttons


viveanban:
**Description**
Currently the zoom buttons are hidden by the bottom cards

**Goals**
- Determine whether we should keep zoom buttons 
- Implement what is decided

**Difficulty (Scale 1-10)**
2

**Additional documentation (Optional)**
![image](https://user-images.githubusercontent.com/32105523/77604345-a0f37800-6ee8-11ea-98ff-1c1f4dd0706a.png)

PR #168, opened 2020-03-26
#154: removed the zoom buttons

Hambrsoom:
I removed the zoom buttons since they haven't been used by the user.


thomasgauvin:
Zoom buttons no longer visible. Good job @Hambrsoom !

90910221_1608232636007193_7033457556534067200_n

Issue #161, opened 2020-03-26
Explore Card and Info-card overlapping


thomasgauvin:
**Describe the bug**
Explore info card and building info card are conflicting/overlapping

**User Story Reference (Optional)**
#11

**To Reproduce**
From map, open building info card.

**Expected behavior**
Info card to overlay on top of explore card.

**Screenshots**
![91057036_294144141564342_5034645172147716096_n.png](https://images.zenhubusercontent.com/5e17455ee6c0ec60e23dc1a0/9634b743-b6a4-41fc-ba8b-0cbd58c13d1a)

thomasgauvin:
Unable to reproduce, reopen if needed

PR #169, opened 2020-03-26
#161/bug fix info card height

tarekait1996:
Fixed the height issue that arose from my last PR.

Please note that the back button currently serves to get the explore card. You won't be able to close the app with the back button. **I was unable to find a solution on how to display the explore card after showing the info card...**

Screenshot_1585249528


thomasgauvin:
Hi Tarek, you can get the explore card back if you press the back button

thomasgauvin:
Hey Tarek,

I don't understand why we cram 2 different fragments within the same UI component. Why not have 2 different UI components so that they don't conflict and can have different behaviors?

Issue #162, opened 2020-03-26
Room markers not showing up on first building selection


thomasgauvin:
**Describe the bug**
The room markers are not showing up when we first select the building.

**User Story Reference (Optional)**
#98 

**To Reproduce**
1. Click building
2. Notice there are no markers, despite the floor being selected

**Expected behavior**
The markers should be indicated, since the floor is already indicated.

**Screenshots**
![91057033_843036902864374_6376466167430643712_n.png](https://images.zenhubusercontent.com/5e17455ee6c0ec60e23dc1a0/207bf2c6-8b3c-4ce8-ba6c-28507925112d)

PR #171, opened 2020-03-26
#162 fixed bug markers not showing by clicking on a building. The mar…

imabhijit:
previously, markers were set to show only when floors are being toggled. This functionality was added to on building click to fix the bug. 

image


viveanban:
Make sure to rebase this over master!

PR #173, opened 2020-03-27
#17: Updated the text displayed to user for next class and added icon

narrapangan:
Expected Behaviour: 
- **If user does not follow format for location:**
the text displayed is "Event is incorrectly formatted"
- **If user enters correct format but class location is not found in the database:**
the text displayed is "Location not found"
- **If user enters correct format and class location is found in the database**
the text displayed is the title of the event and the time until event


narrapangan:
Incorrect Formatting (H-963) = not clickable
Screen Shot 2020-03-27 at 1 09 06 AM

Correct Formatting and Location Not Found in DB (H-9, 912) = not clickable
Screen Shot 2020-03-27 at 1 10 37 AM

Correct Formatting and Found in DB (H-9, 963) = clickable
Screen Shot 2020-03-27 at 1 13 04 AM

narrapangan:
Hey @thomasgauvin ! The text is being updated every time that the search activity is created (therefore, the reason you are probably not getting an update is because you go from calendar straight to the "previous" search activity page). 

By clicking on the back (i.e. returning to the main page) and then clicking on search icon, you will get a new search activity page. 

We could either force the main page to be displayed every time the user exits the app or we could add a refresh button on top of the search activity page. I do not think there's a way we could refresh once the activity has already been created.

PR #178, opened 2020-03-27
#11/POI-Marker: Add POI markers

viveanban:
- Add the 10 first POI markers sorted according to the distance from currentLocation
-  Emphasize on the closest marker by zooming on it
- Open the routes page with pre populated fields when a POI is clicked on.
- Clear the visible POIs when another POI is selected.
- When the user selects a specific floor of a building, only POIs on that floor will be shown.

----General changes----
-  Extracted the code for getting a custom sized icon in a method called getCustomSizedIcon.
- Extracted the code for getting the euclidean distance between coordinates in the Coordinates class. This method will be used when computing the indoor search heuristic and the distance of the POIs.


thomasgauvin:
Would it be possible to remove the markers on back press?

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
```diff
@@ -330,11 +332,15 @@ public boolean setupBuildingMarkerClickListener(GoogleMap map) {
...
             @Override
             public boolean onMarkerClick(Marker marker) {
-                Building building = mViewModel.getBuildingFromeCode(marker.getTag().toString());
-                //TODO: Make function that pops up the info card for the building (via the building-code)
-                String buildingCode = (marker.getTag()).toString();
-                ((MainActivity) getActivity()).showInfoCard(buildingCode);
-                onBuildingClick(building);
+                if (marker.getTag().toString().contains(POI_TAG)) {
+                    //TODO: Use the directions api to get directions to this point of interest
+                    marker.showInfoWindow();
```
thomasgauvin: What does this function do?

viveanban: When you click on a marker, if it's a POI marker, then it should show the directions to that POI.

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
```diff
@@ -330,11 +332,17 @@ public boolean setupBuildingMarkerClickListener(GoogleMap map) {
...
             @Override
             public boolean onMarkerClick(Marker marker) {
-                Building building = mViewModel.getBuildingFromeCode(marker.getTag().toString());
-                //TODO: Make function that pops up the info card for the building (via the building-code)
-                String buildingCode = (marker.getTag()).toString();
-                ((MainActivity) getActivity()).showInfoCard(buildingCode);
-                onBuildingClick(building);
+                if (marker.getTag() != null) {
+                    if (marker.getTag().toString().contains(POI_TAG)) {
+                        //TODO: Use the directions api to get directions to this point of interest
```
thomasgauvin: why do we need to do this? Can't we use the same logic that is already used in the routes activity? Since a POI is a place, we should be able to use the SelectToFromState

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
```diff
@@ -330,11 +332,17 @@ public boolean setupBuildingMarkerClickListener(GoogleMap map) {
...
+                        marker.showInfoWindow();
+                    } else {
+                        Building building = mViewModel.getBuildingFromeCode(marker.getTag().toString());
```
thomasgauvin:
```suggestion
                        Building building = mViewModel.getBuildingFromCode(marker.getTag().toString());
```

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
```diff
@@ -359,22 +367,37 @@ public boolean onMarkerClick(Marker marker) {
...
-                for (WalkingPoint point : walkingPoints) {
-                    LatLng latLng = new LatLng(point.getCoordinate().getLongitude(), point.getCoordinate().getLatitude());
-                    MarkerOptions markerOptions = new MarkerOptions()
-                            .position(latLng)
-                            //TODO: Change the icon to be the same as the POI icon
-                            .icon(BitmapDescriptorFactory.defaultMarker(BitmapDescriptorFactory.HUE_GREEN))
-                            .visible(false);
-                    mMap.addMarker(markerOptions);
+        mViewModel.getListOfPOI().observe(getViewLifecycleOwner(), priorityQueue -> {
+            int position = 1;
```
thomasgauvin: is there a reason why this is a do while and not a for loop?

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -270,14 +263,55 @@ public LatLng getSGWZoomLocation() {
...
+        this.poiList.postValue(getPOIinOrder(allPOI));
+    }
+
+    private PriorityQueue<WalkingPoint> getPOIinOrder(List<WalkingPoint> allPOI) {
+        PriorityQueue<WalkingPoint> orderedList = new PriorityQueue<>((WalkingPoint o1, WalkingPoint o2) -> {
+            Coordinates currentCoordinates = new Coordinates(currentLocation.getLatitude(), currentLocation.getLongitude());
+            double distanceFromO1 = o1.getCoordinate().getEuclideanDistanceFrom(currentCoordinates);
+            double distanceFromO2 = o2.getCoordinate().getEuclideanDistanceFrom(currentCoordinates);
+
+            if (distanceFromO1 < distanceFromO2) return -1;
```
thomasgauvin: can you add a comment to explain -1, 0, 1 pls?

app/src/main/java/com/example/concordia_campus_guide/Models/Coordinates.java
```diff
@@ -45,4 +45,18 @@ public double getLongitude() {
...
         this.longitude = longitude;
     }
+
+    /**
+     * This method returns the euclidean distance between this and another point based on the pythagorean theorem.
+     * @param other
+     * @return
+     */
+    public double getEuclideanDistanceFrom(Coordinates other){
+        final double resultLatDiff = Math.abs(this.toListDouble().get(0)
```
thomasgauvin: any reason why we aren't using the longitude and latitude directly instead of the toListDouble?

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
viveanban: We can only poll a priority queue (heap). It's not easily iterable.

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
viveanban: Yes we can!

viveanban: Nevermind, I got it!

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
thomasgauvin: Ok, can you add a comment to explain this decision? will help understand this code down the line

app/src/main/java/com/example/concordia_campus_guide/Models/Coordinates.java
viveanban: Nope. Will change it.

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
viveanban: Actually, we realized that this can be a for loop. It's just a question of choice.

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -270,14 +261,56 @@ public LatLng getSGWZoomLocation() {
...
+        InputStream deckFile = null;
+        BitmapDescriptor smallMarkerIcon = null;
+        try {
+            deckFile = context.getAssets().open(filename);
+            Bitmap b = BitmapFactory.decodeStream(deckFile);
+            Bitmap smallMarker = Bitmap.createScaledBitmap(b, width, height, false);
+            smallMarkerIcon = BitmapDescriptorFactory.fromBitmap(smallMarker);
+            deckFile.close();
+        } catch (IOException e) {
+            LOGGER.log(Level.WARNING, e.getMessage(), e);
```
ferdousara-parvin: we should be consistent in the logger that we use. For instance, in my DirectionsApiDataParser class, I use the Log object. Did we agree on one?

app/src/main/java/com/example/concordia_campus_guide/Models/Coordinates.java
```diff
@@ -22,6 +22,11 @@ public boolean equals(Object o) {
                 longitude == that.longitude;
     }
 
+    @Override
+    public int hashCode() {
```
ferdousara-parvin: why do we need this?

ferdousara-parvin:
Thanks for the super clear description :) Good job!

app/src/main/java/com/example/concordia_campus_guide/Models/Coordinates.java
viveanban: It was suggested by sonar lint. Basically the reason is that if we are overwriting the equals method, we should also overwrite the hashcode method. This is mostly useful when we use hashmaps.

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
viveanban: I agree. We should come to a conclusion and then change our loggers to reflect that.

ferdousara-parvin: Should we create an improvement task for it?

PR #181, opened 2020-03-28
#94/display gmaps routes ui

ferdousara-parvin:
The last part of #94 is finally completed! It was about adding the onSelected effect on the filter buttons. s/o @thomasgauvin and @Hambrsoom for the help.

image


app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
```diff
@@ -98,24 +98,48 @@ public void onClickFrom(View v){
...
-
+        if(disabilityButton.isSelected())
+            disabilityButton.setSelected(false);
+        else
+            disabilityButton.setSelected(true);
     }
 
     public void onClickShuttle(View v) {
-        getShuttle(v);
+        //getShuttle(v);
```
Hambrsoom: remove it if it isnot needed anymore

ferdousara-parvin: It causes crashes but it is needed. @CarenRizk will have to look into it.

PR #193, opened 2020-03-28
#94/display gmaps routes unit tests

ferdousara-parvin:
@Hambrsoom and I added some unit tests related to US #94


app/src/test/java/com/example/concordia_campus_guide/ModelTests/TrainTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Before
+    public void init() {
+        testUtils = new TestUtils();
+    }
+
+    @Test
+    public void getColorTest() {
+        train = new Train(testUtils.getDirectionStepsObject());
+        assertEquals(testUtils.getDirectionStepsObject().transitDetails.line.shortName,train.getTrainShortName());
```
thomasgauvin: Is this color the shortname? Can you specify this in a comment pls?

app/src/test/java/com/example/concordia_campus_guide/ModelTests/TestUtils.java
```diff
@@ -1,17 +1,23 @@
 package com.example.concordia_campus_guide.ModelTests;
```
thomasgauvin: Can you make this class specific to routes or something by changing the name of this class? Or else all test utils will end up here and that will make this class extremely cluttered

app/src/test/java/com/example/concordia_campus_guide/ModelTests/SubwayTest.java
```diff
@@ -0,0 +1,31 @@
...
+    public void getColorTest() {
+        subway = new Subway(testUtils.getDirectionStepsObject());
+        assertEquals(testUtils.getDirectionStepsObject().transitDetails.line.color,subway.getColor());
+    }
+
+    @Test
+    public void setColorTest(){
+        subway = new Subway();
+        subway.setColor("orange");
+        assertEquals(testUtils.getDirectionStepsObject().transitDetails.line.color,subway.getColor());
```
thomasgauvin: In testing class, we learned that it isn't good practice to have test code be in another file (`mystery guest`)

It would be better to have it in here

app/src/test/java/com/example/concordia_campus_guide/ModelTests/TrainTest.java
ferdousara-parvin: We made a typo here thanks for catching it!

PR #201, opened 2020-03-28
#11: added white background to avoid overlap

tarekait1996:
this should fix the overlap issue


ferdousara-parvin:
Tried it, looks good!

PR #207, opened 2020-03-28
#11/bug fixes

viveanban:
Bugs fixed:

- [x] Fixing erroneous points in walking points
- MB building having 2 entrances (1 tunnel and 1 accessible from the street) was not good design. Only the ground floor entrance is kept; the tunnel entrance is changed to a walking point of type NONE

- [x] Bug Fix in the PathFinder
Fixed the bug where multiple rooms with the same placeCode existed. Rooms do not have only the placeCode as uniqueID, therefore we need to fetch the corresponding WalkingPoint depending on the floorCode and placeCode/roomCode. 

- [x] Handing the case where the current location can be null

- [x] Certain POI are not considered rooms, making the application crash (NPE on displayName bug)

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
@@ -300,7 +300,20 @@ public class LocationFragmentViewModel extends ViewModel {
 
     private PriorityQueue<WalkingPoint> getPOIinOrder(List<WalkingPoint> allPOI) {
         PriorityQueue<WalkingPoint> orderedList = new PriorityQueue<>((WalkingPoint p1, WalkingPoint p2) -> {
-            Coordinates currentCoordinates = new Coordinates(currentLocation.getLatitude(), currentLocation.getLongitude());
+            Coordinates currentCoordinates;
+
+            if (currentLocation == null) {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -300,20 +300,32 @@ public void setListOfPOI(@PoiType String poiType, Context context) {
 
     private PriorityQueue<WalkingPoint> getPOIinOrder(List<WalkingPoint> allPOI) {
         PriorityQueue<WalkingPoint> orderedList = new PriorityQueue<>((WalkingPoint p1, WalkingPoint p2) -> {
-            Coordinates currentCoordinates = new Coordinates(currentLocation.getLatitude(), currentLocation.getLongitude());
-            double distanceFromO1 = p1.getCoordinate().getEuclideanDistanceFrom(currentCoordinates);
-            double distanceFromO2 = p2.getCoordinate().getEuclideanDistanceFrom(currentCoordinates);
+            Coordinates currentCoordinates;
+
+            if (currentLocation == null) {
+                //This building has inverted lat/lng in order to us th geojsons.
```
Hambrsoom: u mean in order to user the geojsons?

ferdousara-parvin: to use the geojsons*

PR #211, opened 2020-03-29
Refactor room markers

thomasgauvin:
this makes the room indicators use markers so that the info card feature can be developed (in the other PR)

not to merge before sprint 4


thomasgauvin:
Merged as part of another pr

Issue #118, opened 2020-03-19
[SONARQUBE BUG]


tarekait1996:
**Describe the bug**
[X] is marked "androidx.annotation.NonNull" but is not initialized in this constructor
Files that are concerned include:
- RoomModel
- WalkingPoint
- Floor
- Building

**Priority**
Low

**Severity**
Low

**Screenshots**

![Screen Shot 2020-03-18 at 8.19.33 PM.png](https://images.zenhubusercontent.com/5e17476e0eda874c7740ff14/809afe36-f100-41ef-b2b7-e90f5e39fe55)
![Screen Shot 2020-03-18 at 8.19.44 PM.png](https://images.zenhubusercontent.com/5e17476e0eda874c7740ff14/8d538ae8-c000-4d31-b531-9206bf0a7532)
![Screen Shot 2020-03-18 at 8.20.00 PM.png]
![Screen Shot 2020-03-18 at 8.20.12 PM.png](https://images.zenhubusercontent.com/5e17476e0eda874c7740ff14/ae6100e2-bac2-4141-9cef-892b0ce4f8db)(https://images.zenhubusercontent.com/5e17476e0eda874c7740ff14/d14ed82d-5655-4c3a-8822-2b12243a29c0)

Hambrsoom:
This bug might be outdated now because there has been some adjustments in the model classes and building doesn't contain an id attribute anymore. As a matter of fact, it gets it from the place class now. Can you check about the status of this bug again n let update us?

PR #212, opened 2020-03-29
#118/codesmells fix V2

tarekait1996:
removed very easy to fix code smells


ferdousara-parvin:
Didn't test it but seems legit. Can someone else just run the app to see if everything is ok?

CarenRizk:
I tried to run the app and looked at the code. Everything looks good :)

Issue #157, opened 2020-03-26
Open Info Card When Classroom Selected In Map


thomasgauvin:
As a user, I would like to be able to select a classroom from the map, so that I can select my destination by selecting it directly from the map.

Acceptance criteria:
Given that I am on the indoor map,
and that I clicked on a classroom,

then I should see,

- A info card pop up at the bottom of the application giving me the option to get directions to that classroom from my current location.

Story tasks:
1. Implement info card on click
2. Implement directions on click of info card button (similar to building info card). This would take the user to the routes page, showing the routes to get to the classroom
3. Testing

thomasgauvin:
I think this will be much more difficult because in the current configuration, Google maps strips all the information from the markers for the rooms. @imabhijit @CarenRizk can you guys check this out?

PR #213, opened 2020-03-29
#157 open info card when classroom selected in map (also for POI)

thomasgauvin:


thomasgauvin:
All tests passing

image

thomasgauvin:
Info card for room

untitled

thomasgauvin:
Info card for POI

untitled

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
```diff
@@ -388,6 +385,35 @@ private void setupPOIListListener() {
...
 
+    private void removeAllRoomsFromMap(){
+        for (Marker marker : roomMarkers) marker.remove();
+        roomMarkers.clear();
+    }
+
+    private void removeAllPOIFromMap(){
+        for (Marker marker : poiMarkers) marker.remove();
+        poiMarkers.clear();
+    }
```
Hambrsoom, line 396: Cann't we create one generic for both methods? Since, it seems that they are serving exactly the  same purpose.

thomasgauvin:
Sorry for not having pr comments with the feature number

app/src/main/res/layout/small_info_card_fragment.xml
```diff
@@ -0,0 +1,46 @@
...
+        android:layout_marginEnd="30sp"
+        android:layout_marginBottom="35sp"
+        android:orientation="horizontal">
+
+        <Button
+            style="@style/InfoCardButton"
+            android:id="@+id/directions"
+            android:layout_marginEnd="15sp"
+            android:drawableLeft="@drawable/ic_directions_black_24dp"
+            android:text="Directions"
```
tarekait1996, line 39: Every text should go in strings.xml, in this case, Directions is written the same way in french but if it was not then we wouldn't be able to get it displayed translated. I am currently going through the whole project and adding the texts to strings.xml and french version of strings.xml. but for new additions, we should already put them there.

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragment.java
thomasgauvin: ok

app/src/main/res/layout/small_info_card_fragment.xml
thomasgauvin: done

PR #214, opened 2020-03-29
Zoom in from when getting directions

thomasgauvin:


app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -60,6 +60,9 @@
     private BitmapDescriptor currentPOIIcon;
     private Location currentLocation;
 
+    //EV centerCoordinates
+    private LatLng initialZoomLocation = new LatLng(45.495638 ,-73.578258);
```
viveanban: Why is this the default? Isn't it supposed to be Hall?

thomasgauvin: default was always EV

viveanban: mmm Do you think that we should put EV as the "default" currentLocation when we calculate distances b/w POIs too?

viveanban:
Can we also change the floor shown when we do this zoom? Currently if the initial location is a floor lower than the topmost floor, you won't see the directions, until you manually change floors.

thomasgauvin:
@viveanban that's more difficult, need to open a different PR

thomasgauvin:
Tests passing, but system tests need to be fixed (which I have done in my other PR)

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
tarekait1996: could we add those to class constant just so it is clear which coordinate we are using?

thomasgauvin: done tarek

thomasgauvin: and Viveka, no, I don't think so

PR #216, opened 2020-03-29
#97: Added several new tests and fixed some

Hambrsoom:
I have implemented some unit and instrumental tests. Our test coverage is 76% now

### DIFF
--- a/app/src/androidTest/java/com/example/concordia_campus_guide/InstrumentalTests/LocationFragmentInstrumentalTest.java
+++ b/app/src/androidTest/java/com/example/concordia_campus_guide/InstrumentalTests/LocationFragmentInstrumentalTest.java
@@ -68,15 +68,6 @@ public class LocationFragmentInstrumentalTest {
         assertNotNull(bitmap2);
     }
 
-//    @Test
-//    public void setAndGetPOIIcon(){
-//        viewModel.setListOfPOI("ELEVATOR",appContext);
-//      //  String as = listOfPOIs.getValue().peek().getPointType().toString();
-//        WalkingPoint walkingPoint = viewModel.getListOfPOI().getValue().peek();
-//        PointType pointType = walkingPoint.getPointType();
-//        assertEquals("ELEVATOR", pointType.toString());
-//    }
-
     @Test
     public void getFloorPolylinesTest(){
         RoomModel roomFrom = new RoomModel(new Coordinates(-73.57907921075821, 45.49702057370776), "823", "H-8");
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/example/concordia_campus_guide/Global/User.java
```diff
@@ -1,30 +1,30 @@
...
-
-    public User() {}
-
-    private void fetchSchedule() {}
-
-    public User getInstance() {
-        return instance;
-    }
-}
+//package com.example.concordia_campus_guide.Global;
```
viveanban: Can we add a TODO and create a task to fix this test?

app/src/androidTest/java/com/example/concordia_campus_guide/InstrumentalTests/LocationFragmentInstrumentalTest.java
```diff
@@ -46,5 +68,38 @@ public void styleMarker(){
         assertNotNull(bitmap2);
     }
 
+//    @Test
```
viveanban: Same as before, can we add a todo and create an issue to fix the test?

CarenRizk: Having this much commented out code will increase our code smells... Is there a way to avoid this Hampic ?

viveanban: For particular tests, we can put the @Ignore annotation in front or delete the @test annotation. Can we do that instead if commenting out things?

app/src/main/java/com/example/concordia_campus_guide/Global/User.java
Hambrsoom: ok this is not a test. It is the user class that we have implemented. I commented it out because i was trying something. I will uncomment it before i merge.

app/src/androidTest/java/com/example/concordia_campus_guide/InstrumentalTests/LocationFragmentInstrumentalTest.java
Hambrsoom: I will put the ignore. and put a todo on it because it is a test we can fix later. Good catch 👍

Hambrsoom: Actually, instead of Todo, I will create a task issue for it.

PR #217, opened 2020-03-29
#7/Bug Fixes for the directions

viveanban:
- [x] Changes to the data concerning the entrances
- Entrance rooms have a place code of entrance
- Changes to the methods needed the entrance, the room datas and walking points data has been made

- [x] Indoor-Outdoor-Indoor Directions fixed
- Refactored the paths class to consider indoor-outdoor-indoor direction cases

- [ ] Zoom in on initial location when displaying directions 
(Implemented by @thomasgauvin , can be cherry-picked to this branch)

- [x] H-8 (only that one 🤔 ) displays an extra polyline from the destination/initial location to the elevators... 
image

~Crashes happen very often (don't know if I should blame my emulator or the code)
Following errors are present (don't know if they are also present on master)~

1. ~E/BUILDING: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference~
2. ~W/System: A resource failed to call close.~
3. ~I/Choreographer: Skipped 125 frames!  The application may be doing too much work on its main thread.~

2 and 3 are probably caused by the emulator.


app/src/main/java/com/example/concordia_campus_guide/Activities/PathsActivity.java
```diff
@@ -130,19 +130,18 @@ public void checkFromToType(Place from, Place to) {
...
+            locationFragment.setIndoorPaths(mViewModel.getEntrance(to), to);
         }
-        if(toIsIndoor && !fromIsIndoor){
-            floorCode = ((RoomModel) to).getFloorCode();
-            tempFrom = new RoomModel(to.getCenterCoordinates(), floorCode.substring(0, floorCode.indexOf('-')), floorCode);
+        else if(mViewModel.arePlacesSeparatedByATunnel(from, to)){
+            locationFragment.setIndoorPaths(from, to);
+        }
+        //[from -> from_entrance ] + outdoor directions + [to_entrance -> to]
+        else if (toIsIndoor && fromIsIndoor) {
```
viveanban: I think adding a check here for when the to-from are from the same building would fix the H-8 weird line problem.

PR #219, opened 2020-03-29
#95 Shuttle Unit test

CarenRizk:
separated some shuttle methods in the routeActivityViewModel to be able to add test cases and added unit tests accordingly

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
@@ -8,9 +8,11 @@ import android.widget.AdapterView;
 import android.widget.ImageButton;
 import android.widget.ListView;
 import android.widget.TextView;
+
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.appcompat.widget.Toolbar;
 import androidx.lifecycle.ViewModelProvider;
+
 import com.example.concordia_campus_guide.Adapters.RoutesAdapter;
 import com.example.concordia_campus_guide.ClassConstants;
 import com.example.concordia_campus_guide.Global.SelectingToFromState;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
```diff
@@ -190,29 +191,29 @@ public void onItemClick(AdapterView<?> adapterView, View view, int i, long l) {
...
+        Intent openSearch = new Intent(RoutesActivity.this, SearchActivity.class);
         startActivity(openSearch);
     }
 
     public void getShuttle(View view) {
-        List<Shuttle> shuttles = mViewModel.getShuttles();
+        List<Shuttle> shuttles = mViewModel.getAllShuttles();
         String shuttleDisplayText = mViewModel.getShuttleDisplayText(shuttles);
-        this.content.setText(shuttleDisplayText);
+        // this.content.setText(shuttleDisplayText);
```
imabhijit: why is this commented out

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -53,24 +55,33 @@ public void setShuttles() {
...
-            }
+        if (from != null && to != null && from.getCampus() != null && to.getCampus() != null) {
+            campusFrom = from.getCampus();
+            campusTo =  to.getCampus();
         }
+        return campusFrom + "," + campusTo;
+    }
 
+    public List<Shuttle> getAllShuttles() {
+        String[] campuses = getFromAndToCampus(getFrom(), getTo()).split(",");
```
imabhijit: you could have also made two strings, using substring but this is fine.

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
CarenRizk: I will change it in a separate PR to be a list so we can display it with the routeOptions.. But good catch I'll remove it from here

PR #220, opened 2020-03-29
#7/draw paths additions

imabhijit:
-Added Colors to the outdoor direction paths, and did some refactoring

image


imabhijit:
Merged my changes with the #217

Issue #51, opened 2020-02-26
Show routes according search and user configuration indoor and outdoor


thomasgauvin:
As a user, I would like to be able to see the different indoor routes that can be taken, so that I can choose my preferred route.

Acceptance criteria:
Given that I chose my starting and ending points,
and I am on the Routes page,
then I should be able to:
- see each possible indoor route
- see a mix of indoor and outdoor routes
- see the time and transit type.

![image](https://user-images.githubusercontent.com/35609369/75819819-45125500-5d69-11ea-9ed9-93bf42860487.png)

Story tasks:
1. Implement getting routes from indoor direction
2. Connect into with GMap API directions (#94 )
3. Connect into with Shuttle directions (#95 )
4. Display routes list of hybrid route options
5. Test

PR #222, opened 2020-03-29
#95/add shuttle route options

CarenRizk:
Incorporated shuttle list with route options
Screenshot_20200329-202808_Concordia-Campus-Guide


thomasgauvin:
does this allow the shuttle to be displayed in the routes page and in the paths page?

thomasgauvin:
I checked out this branch and I am not seeing shuttle in the routes

thomasgauvin:
Screenshot_1585529702

app/src/main/res/raw/rooms_info.json
```diff
@@ -3,6 +3,7 @@
     {
       "roomCode": "VL",
       "floorCode": "VL-1",
+      "Campus": "LOY",
```
ferdousara-parvin: will the fact that campus is with a capital "C" affect anything?

app/src/test/java/com/example/concordia_campus_guide/ViewModel/RoutesActivityViewModelTest.java
```diff
@@ -91,13 +91,17 @@ public void getFromAndToCampuses() {
     }
 
     @Test
-    public void getShuttlesByDayAndTimeTest() {
-        assertEquals(0, mViewModel.getShuttlesByDayAndTime("SGW", "Sunday", 7.05).size());
+    public void adaptShuttleToRoutesTest() {
+        List<Shuttle> shuttles = new ArrayList<>();
+        shuttles.add(shuttle1);
+        mViewModel.adaptShuttleToRoutes(shuttles);
+        mViewModel.getShuttles();
```
ferdousara-parvin: why do you need that line?

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
```diff
@@ -181,11 +184,14 @@ private void setRoutesAdapter() {
         allRoutes.setOnItemClickListener(new AdapterView.OnItemClickListener() {
             @Override
             public void onItemClick(AdapterView<?> adapterView, View view, int i, long l) {
-                Intent openPaths = new Intent(RoutesActivity.this,
```
thomasgauvin: You are disabling the option to go to the paths view for all, not only the shuttle

ferdousara-parvin:
> 
> 
> I checked out this branch and I am not seeing shuttle in the routes

Maybe because it's Sunday? We should probably show a text saying "No shuttle available today" when that happens IF that's the reason.

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -1,43 +1,54 @@
...
+
+import static com.example.concordia_campus_guide.Fragments.LocationFragment.LocationFragmentViewModel.LOGGER;
 
 public class RoutesActivityViewModel extends ViewModel {
 
     private AppDatabase appDB;
     private Place from;
     private Place to;
-    private @ClassConstants.TransportType String transportType = ClassConstants.TRANSIT; // default value
+    private @ClassConstants.TransportType
```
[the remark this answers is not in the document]
CarenRizk: When I do format code it changes it to 2 lines... I think we should stick with it it's a good practice

thomasgauvin:
> > I checked out this branch and I am not seeing shuttle in the routes
> 
> Maybe because it's Sunday? We should probably show a text saying "No shuttle available today" when that happens IF that's the reason.

yes that is the reason, I spoke to Caren and she said she will make it a blank page

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -51,25 +62,21 @@ public void setDirectionsResult(DirectionsResult directionsResult) {
...
         if (from != null && to != null && from.getCampus() != null && to.getCampus() != null) {
             campusFrom = from.getCampus();
-            campusTo =  to.getCampus();
+            campusTo = to.getCampus();
         }
         return campusFrom + "," + campusTo;
     }
 
-    public List<Shuttle> getAllShuttles() {
+    public List<Shuttle> filterShuttles() {
```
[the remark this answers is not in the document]
CarenRizk: get shuttles according to campus, day and time

app/src/main/res/raw/rooms_info.json
CarenRizk: No it is with a capital C in Place.java this is why I had to put in like this.. so it can add it to the place using inheritance

thomasgauvin: I'm not sure it was necessary to add the campus attribute to the rooms object because you can access that from the floor object that you can get using the floor code

thomasgauvin:
> does this allow the shuttle to be displayed in the routes page and in the paths page?

for reference, no, this will be done in #51

PR #224, opened 2020-03-29
#94/unit tests

ferdousara-parvin:
Added more unit tests for #94


app/src/test/java/com/example/concordia_campus_guide/ModelTests/RouteTest.java
```diff
@@ -61,6 +61,17 @@ public void getMainTransportTypeTest() {
...
+    @Test
+    public void getSetSteps() {
+        // Arrange
+        List<TransportType> expectedSteps = new ArrayList<>();
+        Route route = new Route("20 mins", "Autoroute 15-S", "driving");
+        route.setSteps(expectedSteps);
+
+        // Act & Assert
+        assertEquals(expectedSteps, route.getSteps());
+    }
```
tarekait1996: minor suggestion, but usually we add the word "Test" at the end of the method name

ferdousara-parvin: oh yeah I normally do that but forgot to do it fo rthat one, thanks!

app/src/test/java/com/example/concordia_campus_guide/HelpersTest/RoutesHelperTest/DirectionsApiDataParserTest.java
```diff
@@ -112,6 +112,49 @@ public void getDirectionsResultObj_IncorrectDataFormatTest() {
...
+        directionsRoute.summary = "test";
+
+        DirectionsStep d1 = new DirectionsStep();
+        d1.duration = new Duration();
+        d1.duration.value = 142;
+        d1.travelMode = TravelMode.DRIVING;
+
+        DirectionsStep[] steps = {d1};
+
+        directionsRoute.legs[0].steps = steps;
```
tarekait1996: this is kind of a big method, maybe you can break down into a helper that returns the direction route u need, in the future, we might need to use it for another test and it would help to access a method that returns it to you. Or in the init().

Issue #159, opened 2020-03-26
Preferences (for use by accessibility, language, etc.)


thomasgauvin:
As a developer, I would like to be able to create a profile or preferences store, so that I can store user preferences in a state that is maintained once the application is closed

Acceptance criteria:
1. A preferences state/profile/etc. exists, and can be used to store preferences about language, accessibility (these variables will be created as part of other stories)
2. The preference state is kept after closing the application, and can be seen upon re-opening of the application

Story tasks:
1. Research how to implement the profile/state/Android saved instance?
2. Implement it
3. Testing

PR #235, opened 2020-04-03
#159: able to save the state of the disability button on Routes Activity

ferdousara-parvin:
s/o @imabhijit for showing us the SharedPreferences thing on Android Studio!
What I accomplished:
- I was able to save the state of the disability button on the Routes Activity page. 

Recordings coming soon.

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivity.java
@@ -1,6 +1,7 @@
 package com.example.concordia_campus_guide.Activities;
 
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.view.KeyEvent;
 import android.view.View;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -70,6 +74,40 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        super.onResume();
+
+        SharedPreferences sharedPreferences = getSharedPreferences(ClassConstants.SHARED_PREFERENCES, MODE_PRIVATE);
+        for(CompoundButton toggleButton: toggleButtonAndCorrespondingToggleType.keySet()) {
+            String value = sharedPreferences.getString(toggleButtonAndCorrespondingToggleType.get(toggleButton), ClassConstants.FALSE);
+            if(value.equals(ClassConstants.TRUE))
+                toggleButton.setChecked(true);
+            else
+                toggleButton.setChecked(false);
+        }
```
CarenRizk, line 89: SonarLint is suggesting having the if/else in one line can you change it to avoid code smells

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -117,26 +163,29 @@ private void setupSideMenuItemListeners() {
...
-            CompoundButton switchView = (CompoundButton) MenuItemCompat.getActionView(switchItem);
-            setupOnChangeListenerForSwitch(switchView);
-        }
+        MenuItem staffItem = navigationView.getMenu().findItem(R.id.nav_staff);
+        staffToggle = (CompoundButton) staffItem.getActionView();
+        setupOnChangeListenerForSwitch(staffToggle);
+
+        MenuItem accessibilityItem = navigationView.getMenu().findItem(R.id.nav_accessibility);
+        accessibilityToggle = (CompoundButton) accessibilityItem.getActionView();
+        setupOnChangeListenerForSwitch(accessibilityToggle);
```
tarekait1996, line 181: initially, I did the list and for loop to avoid repeating similar lines for each item we have in the nav view, for now, we have 3 toggles which is not a lot but if we had 10 it would have been really repetitive. You could still have a list of your MenuItems and then loop through it, inside the loop you would get a toggle from the menu item and call its onChangeListener.  This is just an idea I had, let me know if you like it or no. Right now it would help remove 4 duplicate lines, it's minor haha

ferdousara-parvin: Yes I completely agree about it being redundant! I removed your loop because I add to use the compound buttons somewhere else in the code but you were creating them only in this portion of the code inside the for loop. That was a quick solution tho, I'll look into your suggestion instead.

app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
ferdousara-parvin: yeah i wasn't sure how they wanted it but thanks Tarek, will make the change!

Issue #153, opened 2020-03-26
"My current location" cannot be set as an origin or a destination on Routes Activity page


ferdousara-parvin:
**Describe the bug**
"My location" is set by default as the "from" in the Routes Activity page. However, when you change the "from" to something else like a building, there is no way to set "My location" anymore. In other words, it is not one of the suggestions in the search results in the Search Activity page

**User Story Reference (Optional)**
No clue but it concerns those who worked on the Search Activity page.

**To Reproduce**
1. Open the application
2. Click on Search button (redirected to Search Activity page)
3. Click on any building (redirected to Routes Activity page)
4. Click on the "from" text field (redirected to Search Activity page)
5. Click on any building (redirected to Routes Activity page)
6. Click on the "from" or "to" text field (redirected to Search Activity page)

There's no way to retrieve "My location" anymore from the search results.

**Expected behavior**
In the search results list or somewhere in the Search Activity page, I should be able to select my current location as an option.

**Screenshots**
None.

PR #238, opened 2020-04-04
#153: Added current location to search results list

narrapangan:
Current location will be added at the top of the list of search results. I found this solution to be better than having a designated button on the search activity page because it will only be displayed if the current location is available. 

Search Activity Page
**Note: Will Show on Map be implemented?**
Screen Shot 2020-04-04 at 8 01 51 AM

Once user taps on "Search Here", the search results (and current location) will be displayed
Screen Shot 2020-04-04 at 8 02 03 AM

Once user taps on "My Location: xxxx, xxxx", the current location is populated 
Screen Shot 2020-04-04 at 8 04 54 AM


ferdousara-parvin:
Can we not show the lat and long tho?

ferdousara-parvin:
Also to answer your question "Will Show on Map be implemented?" --> Answer is no so can you remove it please?

narrapangan:
@ferdousara-parvin It was what was set as default in the my current place. I assumed we wanted it that way, I will change it. And I will remove the Show on Map. 

Thanks!

narrapangan:
Show on Map is removed. (However seems quite plain, any ideas on what we could add?)
image

Coordinates were removed! 
image

app/src/main/java/com/example/concordia_campus_guide/Models/MyCurrentPlace.java
```diff
@@ -6,7 +6,7 @@
 
     public MyCurrentPlace(Double latitude, Double longitude){
         this.setCenterCoordinates(new Coordinates(latitude, longitude));
-        this.displayName = "My location: " + latitude.toString() + ", " + longitude.toString();
+        this.displayName = "My Current Location";
```
tarekait1996, line 9: use string resource, please do not hard code the string anymore

thomasgauvin:
I think it was ok to show the lat lng though

thomasgauvin:
https://github.com/Concordia-Campus-Guide/Concordia-Campus-Guide/pull/242

check this out Narra, this is what I was talking about

narrapangan:
@thomasgauvin I was thinking of doing it like that, however once the user taps on the search bar, there is no way of going back to that initial view.

ferdousara-parvin:
> 
> 
> I think it was better to show the lat lng though

But it's ugly no? Like a real user doesn't really care.

app/src/main/java/com/example/concordia_campus_guide/Models/MyCurrentPlace.java
narrapangan: I updated that (look below)

narrapangan:
image

app/src/main/java/com/example/concordia_campus_guide/Models/MyCurrentPlace.java
```diff
@@ -1,14 +1,20 @@
...
+import android.content.Context;
+
+import com.example.concordia_campus_guide.R;
+
 public class MyCurrentPlace extends Place {
-    String displayName = "Select location";
+    String displayName = "";
 
-    public MyCurrentPlace(Double latitude, Double longitude){
+    public MyCurrentPlace(Context context, Double longitude, Double latitude){
```
narrapangan, line 10: Context had to be added to access string resource

app/src/test/java/com/example/concordia_campus_guide/ModelTests/MyCurrentPlaceTest.java
```diff
@@ -12,13 +12,14 @@
 
     @Before
     public void init() {
-        currentPlace = new MyCurrentPlace(-73.57907921075821, 45.49702057370776);
+        //TODO: add context
```
narrapangan, line 15: Will most likely end up doing instrumented test for this

thomasgauvin:
🔥

Issue #226, opened 2020-03-30
Decrease Number of Code Smells 


ferdousara-parvin:
**Description**
Fix code by looking at suggestions provided by Sonarlint to decrease the number of code smells 

**Goals**
This will help with the quality of the code and will ensure we have better results for our quality metrics.

**Reference to User stories**
None.

**Difficulty (Scale 1-10)**
5

**Additional documentation (Optional)**

PR #239, opened 2020-04-05
#226/decrease code smells

ferdousara-parvin:
@CarenRizk and I fixed some easy to fix code smells.


app/src/main/java/com/example/concordia_campus_guide/Adapters/DirectionWrapper.java
```diff
@@ -32,12 +28,16 @@ public void populateAttributesFromStep(DirectionsStep directionStep){
...
-        if(directionStep.travelMode.toString().equals(ClassConstants.TRANSIT)){
+        if (directionStep.travelMode.toString().equals(ClassConstants.TRANSIT)) {
             this.transitDetails = directionStep.transitDetails;
         }
     }
 
-    private String getTransitType(TravelMode travelMode){
+    public DirectionWrapper() {
+        // being used in tests
+    }
```
tarekait1996, line 38: why isn't the default constructor on top of its class? I think it should be the first method

ferdousara-parvin: I agree w/ you. @CarenRizk do you know why?

imabhijit: no particular reason, i guess we just made it after when we were making the tests

app/src/main/java/com/example/concordia_campus_guide/Database/Converters/CoordinatesListConverter.java
```diff
@@ -22,18 +23,16 @@ public String convertToDatabaseColumn(List<Coordinates> listOfCoordinates) {
...
-        String json = gson.toJson(listOfCoordinates, type);
-        return json;
+        return gson.toJson(listOfCoordinates, type);
     }
 
     @TypeConverter
     public List<Coordinates> convertToEntityAttribute(String listOfCoordinatesString) {
         if (listOfCoordinatesString == null){
-            return null;
+            return new ArrayList<>();
```
imabhijit, line 32: did you change the place where this was being used to check for .empty() instead of null?

tarekait1996:
something I noticed is that most of it were spacing fixes, however, in sonarqube the code smells were not related to spacing or indentation... Are you verifying how much the number decreased?

CarenRizk:
> something I noticed is that most of it were spacing fixes, however, in sonarqube the code smells were not related to spacing or indentation... Are you verifying how much the number decreased?

No we couldn't run sonarqube unfortunately.. The spacing fixes are because of the code formatting I think it looks way cleaner with it but it does not have to do with the code smells

tarekait1996:
sounds good, we should talk to get you guys setup with sonarqube then.

app/src/main/java/com/example/concordia_campus_guide/Database/Converters/CoordinatesListConverter.java
ferdousara-parvin: I think you did @CarenRizk remember I was telling you some tests were failing b/c of that and you fixed them right?

CarenRizk: It is inside the typeConverter so it should be fine

Issue #240, opened 2020-04-05
Fix zoom level for campus switcher buttons (and add all buildings to the SGW campus)


thomasgauvin:
**Description**
![image.png](https://images.zenhubusercontent.com/5e17455ee6c0ec60e23dc1a0/87e54bb6-d924-48c2-81d3-62ac2bdbbf51)

**Goals**
Describe what this proposal will accomplish

**Reference to User stories**
Refer to the user stories that this improvement would affect.

**Difficulty (Scale 1-10)**
On a scale of 1-10, how hard is it to implement this improvement?

**Additional documentation (Optional)**
Links to images, mockups, layouts

PR #241, opened 2020-04-05
#240: added missing building corner coordinates, recentered some buil…

thomasgauvin:
…dings, made the building markers using paint bitmap, removed previous PNGs used for building labels, changed center of Loyola campus to CC building and center of SGW center to EV building, made cmapus switcher zoom level so that we can see the whole campus as per PO request


thomasgauvin:
> Looks really good! The only thing that I might change are attribute names in Class Constants to avoid code smells but it will be fixed as part of #226 so it's fine :)

Ok thanks Karen!

thomasgauvin:
btw, I added a video on slack of what it looks like now (can't upload to github because of file size limit)

app/src/main/java/com/example/concordia_campus_guide/Fragments/LocationFragment/LocationFragmentViewModel.java
```diff
@@ -197,14 +200,27 @@ public void addBuildingMarker(GoogleMap map, LatLng centerPos, String buildingLa
...
+     * The purpose of this method is to create the marker
      *
-     * @param buildingLabel the label of the building
-     * @param context       the context of the LocationFragment
+     * @param label the label of the building
      * @return it will BitmapDescriptor object to use it as an icon for the marker on the map.
      */
-    public BitmapDescriptor styleMarker(String buildingLabel, Context context) {
-        return getCustomSizedIcon("BuildingLabels/" + buildingLabel.toLowerCase() + ".png", context, 150, 150);
+    public Bitmap createBitmapMarkerIcon(String label){
```
imabhijit, line 208: amazing, I've been looking for this all my life

Issue #158, opened 2020-03-26
Translation


thomasgauvin:
As a user, I would like to be able to browse the application in french, so that I can understand how to use the application as a french-speaker.

Acceptance criteria:
Given that I am on the hamburger menu,
and that I toggled the French button

then I should see,

- The menu, step-by-step directions, buttons, and application text written in French
- Data from services and departments, as they are static names from the University, written in English (not translated)

Story tasks:

- [x] Add language preference to side menu

- [x] Change static text based on the preferences

- [x] Change Google Maps directions to French, either there is a Google Maps setting to get directions in French, or we will have to use Google Translate

- [ ] Testing

PR #262, opened 2020-04-08
#158/French translation

tarekait1996:
**This PR looks bigger than it is**

This PR is related to user story #158 which consisted of allowing the user to see french content when switching using the french toggle accessible from the side menu.

How was this achieved?
We created a french version of strings.xml with the translation of every static text in the app. That way based on the locale language, we will access the correct version of each string. we do this by using getResourses.getString(id of the string).

We also updated the URL to get direction so that it shows the right language version of the text

To achieve the language update once the user clicks on the switch, we have to refresh the main activity. Therefore the app blinks for a second, we did not find a way to avoid it while not closing the app.

one small thing we noticed: for the next class in search activity it does not seem to translate it. We spent a lot of time on it and couldn't figure out the issue. So we are making a PR right now to avoid pushing the PR to the last minute.

**WHAT WAS DONE:**
- static text translation in the whole app, this include: explore page, side menu, search page, routes page, calendar pop up and more
- calendar fix by narra
- google map direction translation to french

**WHAT WAS NOT DONE: **
- next class in search page still shows in english even with french toggle turned on


app/src/main/java/com/example/concordia_campus_guide/Activities/SearchActivity.java
```diff
@@ -107,27 +110,40 @@ public void onItemClick(AdapterView<?> parent, View view, int position, long id)
...
+            }else{
+                eventString = calendarViewModel.getNextClassString((calendarEvent));
+                nextClassText.setText(eventString);
+                // double check this value
+                if(eventString.equals(getResources().getString(R.string.incorrect_format_event))){
+                    shouldSetNextClassClickListener = false;
+                }
+                else{
+                    shouldSetNextClassClickListener = true;
+                    nextClassPlace = place;
```
Hambrsoom, line 146: For each if else statement, can you put them into separate functions ? it would make the function cleaner. what do u think ?

app/src/main/java/com/example/concordia_campus_guide/Helper/LocaleHelper.java
```diff
@@ -0,0 +1,65 @@
...
+            lang = getLanguage(context);
+        }
+
+        setLocale(context, lang);
+    }
+
+    public static void onCreate(Context context, String defaultLanguage) {
+        String lang = getPersistedData(context, defaultLanguage);
+        setLocale(context, lang);
+    }
```
Hambrsoom, line 30: LocalHelper is not a fragment or an activity. Why does it have onCreate() methods ?

app/src/main/java/com/example/concordia_campus_guide/Activities/RoutesActivityViewModel.java
```diff
@@ -146,9 +149,9 @@ public void setTransportType(String transportType) {
         this.transportType = transportType;
     }
 
-    public void noShuttles() {
+    public void noShuttles(Context context) {
```
viveanban, line 152: Can this take the string to display as the parameter and avoid taking the context as a param?

viveanban:
Missing translation for when we need to select a location:
image

app/src/main/res/values-fr/strings.xml
```diff
@@ -0,0 +1,56 @@
+<resources>
+    <string name="app_name">Concordia-Campus-Guide</string>
+    <string name="title_activity_maps">Carte</string>
+    <string name="title_activity_routes">RoutesActivity</string>
+    <string name="action_settings">Paramètres</string>
+    <string name="point_of_interest_explore">Explorer</string>
+    <string name="warning_msg">Aucun itinéraire disponible</string>
+    <string name="french">Français</string>
+    <string name="calendar">calendrier google</string>
```
viveanban, line 9: Can this be Calendrier Google? (just to stay consistent with the rest)
image

tarekait1996:
> Missing translation for when we need to select a location:
> image

This is due to the message being in our mode, which we don't have access to our context. without the context, we cannot get the string resources. It might require a refactoring of the show the text for this

Issue #176, opened 2020-03-27
Finish data collection for POI


thomasgauvin:
As a user, I would like to be able to see the different points of interests such as water fountains, lounge areas, so that I can find the points of interest on campus

Acceptance criteria:
Given that I am on the indoor map,
and that I drag the Explore card up,
and that I click on Water Fountains or Lounges,

then I should see,

- The different markers for the POI selected, on the map
- The closest one in a different color to differentiate it from all the other markers

Story tasks:
1. Data collection
2. Show on map
3. Testing
3. ...

PR #264, opened 2020-04-08
#176/data collection

viveanban:
Updated the walking points and the room points to include lounges and water-fountains.


ferdousara-parvin:
Tested it out, looks good :)

Issue #50, opened 2020-02-26
Click from Google Calendar to ConUMaps


thomasgauvin:
As a student, I would like to be able to have access to the ConUMaps application from the google calendar event, so that I can access the application easily.

Acceptance criteria:
Given that I am on my Google Calendar,
and that I click on a class-related event,
and that I click on URL in the event description,
then I should see,

- The ConUMaps application open up
- The Routes Activity page populated with My Location in the origin and with my classroom in the destination
- List of routes that are suggested to me

Story tasks:
1. Integrate with Google Calendar
2. Create task that updates calendar events
3. Test

PR #267, opened 2020-04-08
#50: google calendar permission fixes

Hambrsoom:
The permission for the google calendar will be asked once when the user installs the application for the first time. If the user denies, they can give access to the google calendar through the hamburger menu whenever they want.

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
@@ -40,6 +40,7 @@ import com.example.concordia_campus_guide.Helper.Notification;
 import com.example.concordia_campus_guide.Models.Buildings;
 import com.example.concordia_campus_guide.Models.CalendarEvent;
 import com.example.concordia_campus_guide.Models.Floors;
+import com.example.concordia_campus_guide.Models.Helpers.CalendarViewModel;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/example/concordia_campus_guide/Activities/MainActivity.java
```diff
@@ -422,14 +423,13 @@ public Location getMyCurrentLocation() {
...
-
-            String value = item.isEnabled() ? ClassConstants.TRUE : ClassConstants.FALSE;
-            myEdit.putString(ClassConstants.CALENDAR_INTEGRATION_BUTTON, value);
-
-            myEdit.commit();
+            if(!calendarViewModel.hasReadPermission() && !calendarViewModel.hasWritePermission()){
+                calendarViewModel.askForPermission(this);
+            }
+            else {
+                Toast.makeText(MainActivity.this, "You have already given permission", Toast.LENGTH_SHORT).show();
```
imabhijit, line 431: do you actually need to tell them this?

Hambrsoom: well, so that the user wouldn't get confused if they press on the button and nothing happens

ferdousara-parvin:
Just tested the app! It seems like I am not prompted with the "Access google calendar" popup when i freshly install the app.

Issue #268, opened 2020-04-09
Return indoor direction list in correct order


viveanban:
**Description**
The list of walking points returned by the PathFinder has the destination as the head and the initial point as the tail.

**Goals**
Instead, return a list with the initial point as head and destination as tail.

**Reference to User stories**

**Difficulty (Scale 1-10)**
1

**Additional documentation (Optional)**

PR #270, opened 2020-04-09
#268: Return a solution list with the walking points in order

viveanban:
Initial walking point at the head of a list and destination walking point at the tail of the list.


ferdousara-parvin:
Tested it out, looks good!

Issue #147, opened 2020-03-25
Change the PointType to be a typedef instead of an enum


viveanban:
**Description**
It is a generally a better android programming practice to avoid enums if possible.
> Each value in an ENUM is an object and each declaration will take some runtime memory simply to reference the object. So ENUM values will take more memory then Integer or String constant. 
> Adding a single ENUM will increase the size (13x times than the Integer constant) of final DEX file. It also generates the problem of runtime overhead and your app will required more space.
> So overuse of ENUM in Android would increase DEX size and increase runtime memory allocation size.
([ref](https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3))

**Goals**
In the future, this could help improve performance. For now, this will keep us consistent with how the PoiType is defined. 

**Reference to User stories**
Refer to the user stories that this improvement would affect.

**Difficulty (Scale 1-10)**
3

**Additional documentation (Optional)**

> Before completing this, we will need to change the PointType into a typeDef ([for better performance](https://android.jlelse.eu/android-performance-avoid-using-enum-on-android-326be0794dc3)). The todo is created as a reminder (an improvement issue will probably be created too).

_Originally posted by @viveanban in https://github.com/Concordia-Campus-Guide/Concordia-Campus-Guide/pull/146_

PR #276, opened 2020-04-10
#160/Accessibility : Consider accessibility preferences when finding indoor directions

### LINKED ISSUE #160: Routes Accessibility (handicap, staff)
As a user, I would like to be able to get routes according to my accessibility preferences, so that I can navigate according to my accessibility/handicap/staff.

image

image

Acceptance criteria:
Given that I am on the hamburger menu,
and that I toggled either Accessibility or Staff
OR
Given that I am on the Routes Activity page,
and that I pressed the Disability button,

then I should see,
- The suggested routes change according to my preferences set (handicap, staff). Ex: for handicaps, suggested routes do not include stairs nor escalators but rather elevators.

Story tasks:
1. Add preference toggle in side menu for staff and handicap
2. Change preference/profile according to side menu toggle
3. Add the button in routes page for a one-time selection of handicap routes parameter. In this case, if the user is not handicapped, he can select to make his route accessible without changing their preferences, and can do so directly from the routes page.
4. Change the route suggestions according to preferences

viveanban:
closes #160 
1. Refactor the sharedPreferences to contain boolean values
- Previously, the sharedPreferenced had keys with string values of "true", "false". 
- As it is possible for boolean values to be added to the sharedPreferences, the "true", "false" strings have been replaced by their boolean values
**- Remember to uninstall the application before running this, sinon there will be class conversion errors.**

2. Accessibility pref taken in account for indoor directions
- Shared preferences passed as a parameter to the PathFinder in order to take in account the accessibility pref when coming up with an indoor direction

3. Unit tests refactored to reflect the inclusion of access. pref

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/ClassConstants.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/ClassConstants.java
@@ -18,8 +18,6 @@ import java.util.List;
 public class  ClassConstants {
     public static final String FINE_LOCATION = Manifest.permission.ACCESS_FINE_LOCATION;
     public static final int LOCATION_PERMISSION_REQUEST_CODE = 1234;
-    public static final String TRUE = "true";
-    public static final String FALSE = "false";
     public static final String DISABILITY_BUTTON = "disability_button";
     public static final String CALENDAR_INTEGRATION_BUTTON = "calendar_integration_button";
     public static final String SHARED_PREFERENCES = "UserPreferences";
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/example/concordia_campus_guide/Helper/PathFinder.java
```diff
@@ -117,6 +125,20 @@ public WalkingPoint getWalkingPointCorrespondingToPlace(final Place place) {
         return new ArrayList<>();
     }
 
+    private boolean isWalkingPointAccessible(WalkingPoint point){
+        // TODO: #147 Use string def instead of an enum
```
Hambrsoom, line 129: Is this done?

viveanban: Will be handled in improvement task #147

PR #290, opened 2020-04-10
#226: Removing Code smells

tarekait1996:
fixed code smells such as:

1. Switch to Lambda expressions
2. Abstract class to an interface where possible
3. Renaming of packages to conform with conventions

With these changes. Sonarqube is reporting a total of **53 code smells inside the project,** about 10-15 of those include the renaming of variables which I didn't include as it would require us to make changes in the JSON file too and DB.

I tested the app before opening the pull request, however, I encourage every reviewer to double-check in their end. The changes I made should not affect the behavior of the app, but we never know.


app/src/main/java/com/example/concordia_campus_guide/fragments/InfoCardFragment.java
```diff
@@ -123,10 +123,10 @@ private void setServicesList() {
     }
 
     private void setOnClickListeners(){
-        this.directionsBt.setOnClickListener(v -> onClickDirections(v));
+        this.directionsBt.setOnClickListener(v -> onClickDirections());
     }
 
-    public void onClickDirections(View v){
+    public void onClickDirections(){
```
ferdousara-parvin, line 129: you sure removing the view doesn't affect anything? There were some other places where it was saying to remove it b/c it wasn't being used and when I did, the app wasn't working properly

app/src/main/java/com/example/concordia_campus_guide/models/routes/TransportType.java
```diff
@@ -1,4 +1,4 @@
 package com.example.concordia_campus_guide.models.routes;
 
-public abstract class TransportType {
+public interface TransportType {
```
ferdousara-parvin, line 3: i'm using "instanceof" in a class, don't remember where. Using interface will affect that part of the code no?

viveanban: @ferdousara-parvin Did it affect it when you ran the app?

viveanban: Wait, why is this interface not deleted? It has nothing in it ...

viveanban: Ok I got my answer after a quick search: https://en.wikipedia.org/wiki/Marker_interface_pattern

app/src/main/java/com/example/concordia_campus_guide/fragments/InfoCardFragment.java
tarekait1996: the reason why the one you touched caused trouble is because they are directly linked to the ui actions. The buttons were not setup in the code but in the UI and were given name of function to look for in the code. Therefore the functions have to conform to the standard of including the view in the argument. This one however does not seem to be linked to a UI element from the xml and should be safe. You can double check

app/src/main/java/com/example/concordia_campus_guide/models/routes/TransportType.java
tarekait1996: "The java “instanceof” operator is used to test whether the object is an instance of the specified type (class or subclass or interface)."

tarekait1996: I did not delete it because it is implemented by multiple types of transport like Bus Walk etc...

PR #293, opened 2020-04-10
#284/Floor-Directions: Refactored indoor and outdoor directions to consider floors.

### LINKED ISSUE #284: Direction from/to a place that is a Floor instance returns no routes.
**Describe the bug**
Direction from/to a place that is a Floor instance returns no routes.

**To Reproduce**
Search for directions from H-9 to MB-S2. No directions are given.

**Expected behavior**
Return at indoor directions from the access points of each floor.

**Screenshots**
image

viveanban:
- Previous implementation assumed that indoor directions could only be created for rooms.
- These changes consider Floors set as initial or destination to also be qualified to be indoor directions.
- floors in the database were missing crucial info such as Campus and center-coordinated (these are added as part of this PR).

Closes #284


viveanban:
> Tested the branch nothing is breaking seems good to me :) But why are the icons on the pathsInfoCard black ? I think it looked better in red

image
You mean these right? I haven't changed anything related to their colours tho.

CarenRizk:
> > Tested the branch nothing is breaking seems good to me :) But why are the icons on the pathsInfoCard black ? I think it looked better in red
> 
> image
> You mean these right? I haven't changed anything related to their colours tho.

Yeah I do mean these. Ummm ok it's fine then

PR #295, opened 2020-04-10
#256: Improved coupling for main activity, database, class constants,…

thomasgauvin:
… and many more


thomasgauvin:
Before:
image
After:
image

app/src/main/java/com/example/concordia_campus_guide/ClassConstants.java
```diff
@@ -16,8 +12,6 @@
 import java.util.List;
 
 public class  ClassConstants {
-    public static final String FINE_LOCATION = Manifest.permission.ACCESS_FINE_LOCATION;
-    public static final int LOCATION_PERMISSION_REQUEST_CODE = 1234;
     public static boolean MY_LOCATION_PERMISSION_GRANTED = false;
     public static final String DISABILITY_BUTTON = "disability_button";
     public static final String CALENDAR_INTEGRATION_BUTTON = "calendar_integration_button";
```
Hambrsoom, line 17: Where is this used ?

app/src/main/java/com/example/concordia_campus_guide/ClassConstants.java
```diff
@@ -46,27 +40,14 @@
...
     public static final String SGW_CENTER_BUILDING_LABEL = "EV";
-
     //Loyola building label string used as campus center
     public static final String LOYOLA_CENTER_BUILDING_LABEL = "CC";
 
-    public static final int EARTH_RADIUS_KM = 6371;
-
     public static final String LOCATION_FRAGMENT_TAG = "LocationFragment";
     public static final String POI_TAG = "POI";
     public static final String ROOM_TAG = "ROOM";
```
Hambrsoom, line 53: Those three variables are used in Logger should we keep them here ?

app/src/main/java/com/example/concordia_campus_guide/models/WalkingPoint.java
```diff
@@ -87,6 +88,22 @@ public Coordinates getCoordinate() {
...
+        return coordinate.getLatLng();
+    }
+
+    public double getLatitude(){
+        return coordinate.getLatitude();
+    }
+
+    public double getLongitude(){
+        return coordinate.getLongitude();
+    }
```
Hambrsoom, line 105: Do we need both ? why don't we keep only the method that return LatLng ?

app/src/main/java/com/example/concordia_campus_guide/view_models/PathsViewModel.java
```diff
@@ -18,6 +17,8 @@
 
 public class PathsViewModel extends ViewModel {
 
+    public static final int EARTH_RADIUS_KM = 6371;
```
Hambrsoom, line 20: this is only used in one method no ? so why do we have  an attribute for it ? Wouldn't be better if we just keep inside the getDistanceFromLatLonInKm() method ?

thomasgauvin: I think that constants should be as an attribute of the class where it is used

app/src/main/java/com/example/concordia_campus_guide/models/WalkingPoint.java
thomasgauvin: You need all of these because this ensures that the `LatLng` is not coupled to the class calling `getLongitude` or `getLatitude`

app/src/main/java/com/example/concordia_campus_guide/ClassConstants.java
thomasgauvin: These were reused in many classes, so I thought it was ok to leave them here

app/src/main/java/com/example/concordia_campus_guide/ClassConstants.java
thomasgauvin: app/src/main/java/com/example/concordia_campus_guide/Helper/CurrentLocationPermissionRequest.java

app/src/main/java/com/example/concordia_campus_guide/activities/PathsActivity.java
```diff
@@ -76,8 +80,8 @@ private void initComponents() {
         getBundleInformation();
         setBackButtonOnClickListener();
 
-        locationFragment.setDifferentInitialView(new LatLng(initialLocation.getCenterCoordinates().getLatitude(),
-                initialLocation.getCenterCoordinates().getLongitude()));
+        locationFragment.setDifferentInitialView(new LatLng(initialLocation.getLatitude(),
```
ferdousara-parvin, line 83: this causes me issues when running the app

ferdousara-parvin:
I'm having issues when I want to draw the path after putting in the "from" and "to". I get the following error: 
image

app/src/main/java/com/example/concordia_campus_guide/activities/PathsActivity.java
ferdousara-parvin: it is linked to the error I mentioned in your PR.

thomasgauvin: ok fixed it thanks

thomasgauvin:
> The application is crashing on my side to when any from and to places

Fixed now

app/src/main/java/com/example/concordia_campus_guide/ClassConstants.java
Hambrsoom: Then, maybe we can just keep it in there since it is not used anywhere else too

app/src/main/java/com/example/concordia_campus_guide/view_models/PathsViewModel.java
Hambrsoom: I don't see why? since it is not used any other method in the class

ferdousara-parvin:
Tested the app and it looks fine now!

Issue #179, opened 2020-03-27
Implement a Builder Pattern for Route object


ferdousara-parvin:
**Description**
There's several ways of creating a Route object depending on some factors. 

**Goals**
A Builder design pattern would be ideal for this situation

**Reference to User stories**
Not a particular user story but the focus is the Route class and all classes that create a new Route object.

**Difficulty (Scale 1-10)**
3

**Additional documentation (Optional)**
Links to images, mockups, layouts

PR #297, opened 2020-04-11
#179: implemented builder pattern

ferdousara-parvin:
Created a RouteBuilder class and a RouteBuilderDirector class to implement the Builder design pattern.

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/models/helpers/RouteBuilder.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/models/helpers/RouteBuilder.java
@@ -1,0 +1,46 @@
+package com.example.concordia_campus_guide.models.helpers;
+
+import com.example.concordia_campus_guide.ClassConstants;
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/example/concordia_campus_guide/models/helpers/RouteBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.List;
+
+public class RouteBuilder {
+    private List<TransportType> steps;
+    private String departureTime;
+    private String arrivalTime;
+    private String duration;
+    private String summary;
+    private @ClassConstants.TransportType
+    String mainTransportType;
```
Hambrsoom, line 16: shouldn't be private as well ?

ferdousara-parvin: yeah it's the line on top

PR #299, opened 2020-04-11
286/bug fix route options

CarenRizk:
- Fixed bug when there are no routes available showing the same text for all transport options
- Added system test for US #51 
- Fixed bug when clicking on no routes available in routesActivity the app was breaking

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
@@ -105,8 +105,8 @@ public class RoutesAdapter extends ArrayAdapter<Route> {
     }
 
     private void setUIWalk(Route route) {
-        if (!(route.getSummary().equals("Pas de routes disponible, veuillez changer les critères.") ||
-                route.getSummary().equals("No routes available, please change the filter criteria."))) {
+
+        if (!route.getSummary().equals(context.getResources().getString(R.string.no_routes))) {
             mainTransportType.setImageResource(R.drawable.ic_directions_walking);
         }
         TextView summary = new TextView(context);
@@ -118,8 +118,7 @@ public class RoutesAdapter extends ArrayAdapter<Route> {
     }
 
     private void setUIDriving(Route route) {
-        if (!(route.getSummary().equals("Pas de routes disponible, veuillez changer les critères.") ||
-                route.getSummary().equals("No routes available, please change the filter criteria."))) {
+        if (!route.getSummary().equals(context.getResources().getString(R.string.no_routes))) {
             mainTransportType.setImageResource(R.drawable.ic_directions_driving);
         }
 
@@ -132,8 +131,7 @@ public class RoutesAdapter extends ArrayAdapter<Route> {
     }
 
     private void setUITransit(Route route) {
-        if (route.getSummary().equals("Pas de routes disponible, veuillez changer les critères.") ||
-                route.getSummary().equals("No routes available, please change the filter criteria.")) {
+        if (route.getSummary().equals(context.getResources().getString(R.string.no_routes))) {
             TextView summary = new TextView(context);
             summary.setText(route.getSummary());
             summary.setGravity(Gravity.CENTER);
--- a/app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
@@ -105,7 +105,7 @@ public class RoutesAdapter extends ArrayAdapter<Route> {
     }
 
     private void setUIWalk(Route route) {
-        if (!route.getSteps().isEmpty()) {
+        if (route.getSteps() != null) {
             mainTransportType.setImageResource(R.drawable.ic_directions_walking);
         }
         TextView summary = new TextView(context);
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
```diff
@@ -113,7 +117,9 @@ private void setUIWalk(Route route) {
     }
 
     private void setUIDriving(Route route) {
-        mainTransportType.setImageResource(R.drawable.ic_directions_driving);
+        if (route.getDuration() != null) {
```
ferdousara-parvin, line 120: why specifically if .getDuration() is not null? What about .getSummary() ?

app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
```diff
@@ -102,8 +105,9 @@ private void setUIElements(Route route) {
     }
 
     private void setUIWalk(Route route) {
-        mainTransportType.setImageResource(R.drawable.ic_directions_walking);
-
+        if (route.getSteps() != null) {
```
ferdousara-parvin, line 108: why specifically if route.getSteps() is not null?

app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
```diff
@@ -173,24 +186,25 @@ private void setUIShuttle(Route route) {
...
         durationWalk.setGravity(Gravity.CENTER);
 
-        details.addView(walkIcon);
         details.addView(durationWalk);
     }
 
     private void setUIBus(Bus bus) {
-        ImageView busIcon = new ImageView(context);
-        busIcon.setImageResource(R.drawable.ic_directions_bus_red);
+        if (bus.getBusNumber() != null) {
```
ferdousara-parvin, line 198: why this null check in particular?

app/src/main/java/com/example/concordia_campus_guide/view_models/RoutesActivityViewModel.java
```diff
@@ -151,10 +151,11 @@ public void setTransportType(String transportType) {
         this.transportType = transportType;
     }
 
-    public void noShuttles(String noShuttleText) {
+    public void noRoutesOptions(String noRoutes) {
         RouteBuilder routeBuilder = new RouteBuilder();
-        director.constructShuttleRouteWithSummaryOnly(routeBuilder, noShuttleText);
-        Route shuttleRoute = routeBuilder.getRoute();
-        routeOptions.add(shuttleRoute);
+        routeBuilder.setMainTransportType(transportType);
```
ferdousara-parvin, line 156: Can you use the director instead of using the builder directly. That's the proper chain of commands for the builder design pattern.

ferdousara-parvin:
I tried the app and it seems there's issues such as the following: 
image

app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
CarenRizk: Good point I realized that if we're checking transit we don't need to check the bus

app/src/main/java/com/example/concordia_campus_guide/adapters/RoutesAdapter.java
CarenRizk: I changed all the checks to check for the summary so it's more consistent and the app is behaving correctly now

app/src/main/java/com/example/concordia_campus_guide/view_models/RoutesActivityViewModel.java
CarenRizk: @imabhijit can you take a look at this ?

imabhijit: I'm not sure if that is necessary, this is pretty much the only place where we would need to make a Route without anything besides the summary. This route is not "shuttle-only" anymore, it is for any transportation type.

imabhijit: anyways, I changed it. Let's follow the code standard!

app/src/main/java/com/example/concordia_campus_guide/models/helpers/RouteBuilderDirector.java
```diff
@@ -51,12 +51,8 @@ public void constructShuttleRoute(RouteBuilder builder, String departureTime, St
...
 
-    public void constructShuttleRouteWithSummaryOnly(RouteBuilder builder, String summary) {
-        builder.setSteps(null);
-        builder.setDepartureTime("");
-        builder.setArrivalTime("");
-        builder.setDuration("");
+    public void constructRouteWithSummaryOnly(RouteBuilder builder, String summary, String transportType) {
         builder.setSummary(summary);
-        builder.setMainTransportType(ClassConstants.SHUTTLE);
+        builder.setMainTransportType(transportType);
```
ferdousara-parvin, line 56: shouldn't you set all the fields tho?

ferdousara-parvin: @imabhijit

imabhijit: we want the other fields to be null

Issue #304, opened 2020-04-12
Decrease Vulnerabilities inside the project 


tarekait1996:
**Description**
Those are the current vulnerabilities inside the ConUMap project

![Screen Shot 2020-04-12 at 10 15 50 AM](https://user-images.githubusercontent.com/34817695/79071231-a0a8fa00-7ca8-11ea-842e-5db1a13c3c7f.png)

**Goals**
We should decrease this number at least by one half (4)

PR #305, opened 2020-04-12
#304, #226: Vulnerabilities and Code Smells fix

tarekait1996:
Screen Shot 2020-04-12 at 11 12 21 AM
As we can see in the image, I removed all vulnerabilities and some code smells.


app/src/main/java/com/example/concordia_campus_guide/models/Time.java
```diff
@@ -3,7 +3,31 @@
...
 
 public class Time implements Serializable {
-    public String text;
-    public String timeZone;
-    public long value;
+    private String text;
+    private String timeZone;
+    private long value;
+
+    public String getText() {
```
ferdousara-parvin, line 10: o we have unit tests for it?

tarekait1996: added

PR #307, opened 2020-04-12
#160 reduced image height so it won't compare the street name only th…

CarenRizk:
…e floor plans


thomasgauvin:
I'm glad we got it working on yours too now! 😄

PR #317, opened 2020-04-14
Fixing a walking point to give correct directions

viveanban:
Fixing a walking point so that it gives good directions:
Screenshot_1586835110
image


ferdousara-parvin:
Thanks for fixing it so rapidly!

PR #318, opened 2020-04-14
added a toast when the current location isnot working

Hambrsoom:
if the user clicks on my current location from the search page and the user's current location is not available then display a toast.

### DIFF
--- a/app/src/main/java/com/example/concordia_campus_guide/activities/MainActivity.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/activities/MainActivity.java
@@ -100,8 +100,8 @@ public class MainActivity extends AppCompatActivity implements NavigationView.On
             RoomModel room = appDb.roomDao().getRoomByRoomCodeAndFloorCode(roomCode, floorCode);
 
             if(room!=null){
-                SelectingToFromState.setMyCurrentLocation(currentLocation.getMyLocation());
- //               openRoutesPage(room, this);
+                SelectingToFromState.setMyCurrentLocation(getMyCurrentLocation());
+                openRoutesPage(room, this);
             }
         }
         catch(Exception e){
--- a/app/src/main/java/com/example/concordia_campus_guide/activities/MainActivity.java
+++ b/app/src/main/java/com/example/concordia_campus_guide/activities/MainActivity.java
@@ -100,8 +100,8 @@ public class MainActivity extends AppCompatActivity implements NavigationView.On
             RoomModel room = appDb.roomDao().getRoomByRoomCodeAndFloorCode(roomCode, floorCode);
 
             if(room!=null){
-                SelectingToFromState.setMyCurrentLocation(getMyCurrentLocation());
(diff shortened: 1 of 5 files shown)


viveanban:
Should the toast also be displayed when the location is not available when we click on the icon on the map?
image

app/src/main/java/com/example/concordia_campus_guide/activities/MainActivity.java
```diff
@@ -100,8 +100,8 @@ private void acceptOpenFromCalendarIntent() {
             RoomModel room = appDb.roomDao().getRoomByRoomCodeAndFloorCode(roomCode, floorCode);
 
             if(room!=null){
-                SelectingToFromState.setMyCurrentLocation(getMyCurrentLocation());
-                openRoutesPage(room, this);
+                SelectingToFromState.setMyCurrentLocation(currentLocation.getMyLocation());
+ //               openRoutesPage(room, this);
```
viveanban, line 104: why commented out?

app/src/main/java/com/example/concordia_campus_guide/activities/MainActivity.java
```diff
@@ -100,8 +100,8 @@ private void acceptOpenFromCalendarIntent() {
             RoomModel room = appDb.roomDao().getRoomByRoomCodeAndFloorCode(roomCode, floorCode);
 
             if(room!=null){
-                SelectingToFromState.setMyCurrentLocation(getMyCurrentLocation());
```
thomasgauvin, line 103: This change doesn't do anything

thomasgauvin: getMyCurrentLocation() was already doing that

app/src/main/java/com/example/concordia_campus_guide/helper/CurrentLocation.java
```diff
@@ -45,9 +45,9 @@ public void run() {
                         setMyLocation(location);
                     }
                 });
-                handler.postDelayed(this, 2000);
+                handler.postDelayed(this, 500);
```
thomasgauvin, line 48: Has anyone checked to see if this has performance issues?

app/src/main/java/com/example/concordia_campus_guide/activities/MainActivity.java
Hambrsoom: I will put it back this here good catch

app/src/main/java/com/example/concordia_campus_guide/helper/CurrentLocation.java
Hambrsoom: When I tired it, there wasn't any issues. I asked Viveka and Ferdousara to try it too

viveanban: I tested out the app on an emulator and I am not having any performance issues.

PR #319, opened 2020-04-14
Added missing walking points for rooms 927.02 and 927.03

viveanban:
- App used to crash when we tried getting directions to those classes because a corresponding walking point did not exist

- Changed the connections of pt 220 (for room 927.01)
- Added 2 walking points (814,815) correspondig to rooms (927.02,927.03)
- The rest is just formatting changes [whitespace]

--------------------------
Fixed directions b/w h927-h919
image
image
--------------------------
Fixed directions from H-827 -> VL1-126 where the directions took the stairs from the 8th to then take the elevators on the 9th.
image

### DIFF
--- a/app/src/main/res/raw/walking_points.json
+++ b/app/src/main/res/raw/walking_points.json
@@ -4540,7 +4540,6 @@
       "pointType": "STAIRS",
       "connectedPointsId": [
         410,
-        411,
         217
       ],
       "placeCode": "stairs-3"
@@ -4814,8 +4813,7 @@
       "connectedPointsId": [
         69,
         70,
-        412,
-        387
+        412
       ],
       "placeCode": ""
     },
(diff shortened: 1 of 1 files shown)


app/src/main/res/raw/walking_points.json
```diff
@@ -2795,7 +2795,9 @@
       "floorCode": "H-9",
       "pointType": "CLASSROOM",
       "connectedPointsId": [
```
viveanban, line 2797: Actual change 1

app/src/main/res/raw/walking_points.json
```diff
@@ -6728,6 +7073,32 @@
         54
       ],
       "placeCode": "lounge-2"
+    },
+    {
+      "id": 814,
+      "coordinate": {
+        "longitude": 45.49707321605269,
+        "latitude": -73.57925456017254
+      },
```
viveanban, line 7082: Actual change 2

PR #320, opened 2020-04-14
Added missing point types for infoCard

CarenRizk:
Realized some points have types that are not checked when adding them to the info card and fixed info card icon colors 
image


app/src/main/java/com/example/concordia_campus_guide/fragments/PathInfoCardFragment.java
```diff
@@ -106,7 +106,7 @@ public void createDirectionImageButton(final LinearLayout layout, final LinearLa
         final ImageButton btn = new ImageButton(getContext());
         btn.setLayoutParams(layoutParams);
         btn.setImageResource(imageId);
-        btn.setColorFilter(R.color.colorAppTheme);
+        btn.setColorFilter(getResources().getColor(R.color.colorAppTheme));
```
viveanban, line 109: isn't this deprecated? You can use the getColor from the activity instead.

viveanban, line 109:
```suggestion
        btn.setColorFilter(getActivity().getColor(R.color.colorAppTheme));
```

app/src/main/java/com/example/concordia_campus_guide/view_models/PathsViewModel.java
```diff
@@ -143,6 +141,21 @@ private void addIndoorDescriptionToList(WalkingPoint startWalkingPoint, WalkingP
                     addCardToList("Walk towards classroom " + endWalkingPoint.getPlaceCode(), "Classroom");
                 }
                 break;
+            case PointType.WASHROOM:
```
viveanban, line 144: Is it normal that these are hardcoded? Does it mean it won't be translated?

app/src/main/res/drawable/ic_directions_washroom.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
viveanban, line 1: Can we use a different washroom icon? like a toilet or something more descriptive?

CarenRizk: I couldn't find any :/

app/src/main/java/com/example/concordia_campus_guide/view_models/PathsViewModel.java
CarenRizk: That's true! I'll add them to the resources but I need someone to translate them for me :P

viveanban: Perfect! I'll help you out!

app/src/main/res/drawable/ic_directions_washroom.xml
viveanban: How about wc ? (through the android clipart thing)

image

CarenRizk: Ohhhh didn't see this yeah I'll replace it !

PR #322, opened 2020-04-15
If there are no departure and arrival time there was a dash being dis…

CarenRizk:
…played. This PR removes the dash
image


ferdousara-parvin:
Thanks for fixing it :)

PR #323, opened 2020-04-15
removed indoor map button

ferdousara-parvin:


app/src/main/res/layout/info_card_fragment.xml
```diff
@@ -32,22 +32,14 @@
         android:orientation="horizontal">
 
         <Button
-            style="@style/InfoCardButton"
             android:id="@+id/directions"
+            style="@style/InfoCardButton"
+            android:layout_width="184dp"
```
tarekait1996, line 37: this will not look the same on every phone since you setting a hard coded value btw. you should just set width to wrap content and put it centered

PR #324, opened 2020-04-15
fixed overlap issue with info cards

tarekait1996:


ferdousara-parvin:
will that affect system tests?

